PR #75, opened 2014-02-03
Add assertRaisesRegexp method.

jvarlamova:
The method assertRaisesRegexp was introduced in python 2.7.
This patch implements assertRaisesReagexp method to make possible
using it in python 2.6.

assertRaisesRegexp asserts that a method raises an exception
and given pattern matches on the string representation of the raised
exception.


jml:
Now merged in #185. Thanks so much, and sorry for the delay.

PR #127, opened 2015-01-09
Collect all duplicate test id's before raise

toabctl:
To simplify debugging duplicate test id's, collect all duplicates,
remember the number of duplicates and raise an Exception after all
duplicates are found.

Closes-Bug: #1390082


jml:
Now merged in #187. Thanks so much, and sorry for the delay!

Issue #111, opened 2014-10-16
Hard to tell what went wrong when a text content object is broken


thomir:
I see lots of stack traces such as this one:

```
Traceback

Traceback (most recent call last):
  File "/usr/bin/autopilot3", line 9, in <module>
    load_entry_point('autopilot==1.5.0', 'console_scripts', 'autopilot3')()
  File "/usr/lib/python3/dist-packages/autopilot/run.py", line 729, in main
    test_app.run()
  File "/usr/lib/python3/dist-packages/autopilot/run.py", line 617, in run
    action()
  File "/usr/lib/python3/dist-packages/autopilot/run.py", line 681, in run_tests
    test_result = test_suite.run(result)
  File "/usr/lib/python3.4/unittest/suite.py", line 125, in run
    test(result)
  File "/usr/lib/python3.4/unittest/case.py", line 625, in __call__
    return self.run(*args, **kwds)
  File "/usr/lib/python3/dist-packages/testscenarios/testcase.py", line 62, in run
    test.run(result)
  File "/usr/lib/python3/dist-packages/testscenarios/testcase.py", line 65, in run
    return super(WithScenarios, self).run(result)
  File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 585, in run
    return self.__RunTest(self, self.exception_handlers).run(result)
  File "/usr/lib/python3/dist-packages/testtools/runtest.py", line 74, in run
    return self._run_one(actual_result)
  File "/usr/lib/python3/dist-packages/testtools/runtest.py", line 88, in _run_one
    return self._run_prepared_result(ExtendedToOriginalDecorator(result))
  File "/usr/lib/python3/dist-packages/testtools/runtest.py", line 107, in _run_prepared_result
    handler(self.case, self.result, e)
  File "/usr/lib/python3/dist-packages/testtools/testcase.py", line 558, in _report_failure
    result.addFailure(self, details=self.getDetails())
  File "/usr/lib/python3/dist-packages/testtools/testresult/real.py", line 1149, in addFailure
    return self.decorated.addFailure(test, err)
  File "/usr/lib/python3/dist-packages/autopilot/testresult.py", line 78, in addFailure
    return super(type(self), self).addFailure(test, err, details)
  File "/usr/lib/python3/dist-packages/testtools/testresult/real.py", line 1546, in addFailure
    return self.decorated.addFailure(test, err, details=details)
[...]

PR #129, opened 2015-03-09
Overhaul exception reporting.

rbtcollins:
unittest2 recently added the ability to show local variables in
tracebacks as https://github.com/testing-cabal/testtools/issues/111
requested for us. Reusing that requires some refactoring of our code,
in particular where we were reimplementing bits of the traceback
module. Now we can just hard-depend on traceback2 and linecache2 which
are brought in by unittest2 1.0.0.

Change-Id: Ieb3268029d26b48ed4fcd25ed644bd339f6aa3fb


NEWS
```diff
@@ -7,6 +7,15 @@ Changes and improvements to testtools_, grouped by release.
 NEXT
 ~~~~
 
+Improvements
+------------
+
+* ``testtools.run`` now accepts ``--locals`` to show local variables
+  in tracebacks, which can be a significant aid in debugging. In doing
+  so we've removed the code reimplementing linecache and traceback as
+  the traceback2 and linecache2 modules can now be used instead.
```
jml: I'd call out the new dependency on `traceback2` as a separate bullet point (although, thank heaven that deps aren't quite as much of a pain these days).

Also, nitpick, AIUI, it's not they _can_ be used instead, but rather that we _are_ using them. Is that right?

doc/for-framework-folk.rst
```diff
@@ -92,6 +92,13 @@ cause ``testtools.RunTest`` to fail the test case after the test has finished.
 This is useful when you want to cause a test to fail, but don't want to
 prevent the remainder of the test code from being executed.
 
+Exception formatting
+--------------------
+
+Testtools ``TestCase`` do their own formatting of exceptions. The attribute
```
jml: I think you need the word "instances" after `TestCase`, since it looks singular.

Perhaps, "`TestCase` _instances_ format their own exceptions"?

testtools/compat.py
```diff
@@ -26,10 +25,12 @@
 import traceback
 import unicodedata
 
-from extras import try_imports
+from extras import try_import, try_imports
 
 BytesIO = try_imports(['StringIO.StringIO', 'io.BytesIO'])
 StringIO = try_imports(['StringIO.StringIO', 'io.StringIO'])
+# To let setup.py work, make this a conditional import.
```
jml: What does that mean exactly? What does `setup.py` do if it is not present? Do we have to do this for all of our external dependencies?

testtools/content.py
```diff
@@ -163,61 +161,52 @@ def __init__(self, stack_lines, prefix_content="", postfix_content=""):
     def _stack_lines_to_unicode(self, stack_lines):
         """Converts a list of pre-processed stack lines into a unicode string.
         """
+        msg_lines = traceback.format_list(stack_lines)
```
jml: I guess the conditional import isn't really meant to be conditional then.

testtools/content.py
```diff
@@ -163,61 +161,52 @@ def __init__(self, stack_lines, prefix_content="", postfix_content=""):
...
-
-    return StackLinesContent(stack_lines, prefix, postfix)
+        :param err: An exc_info error tuple.
+        :param test: A test object used to obtain failureException.
+        :param capture_locals: If true, show locals in the traceback.
+        """
+        if err is None:
+            raise ValueError("err may not be None")
+
+        exctype, value, tb = err
```
jml: Why handle the `err is None` case specially? Tuple unpacking will raise a perfectly cromulent `TypeError` anyway.

testtools/content.py
```diff
@@ -232,22 +221,20 @@ def StacktraceContent(prefix_content="", postfix_content=""):
...
-    frames_only = [line[0] for line in stack[:limit]]
-    processed_stack = [ ]
-    for frame in reversed(frames_only):
-        filename, line, function, context, _ = inspect.getframeinfo(frame)
-        context = ''.join(context)
-        processed_stack.append((filename, line, function, context))
-    return StackLinesContent(processed_stack, prefix_content, postfix_content)
+    stack = traceback.walk_stack(None)
+    def filter_stack(stack):
+        # Discard this frame.
```
jml: s/this/the filter_stack frame/

jml:
The [pypy3 tests fail](https://travis-ci.org/testing-cabal/testtools/jobs/53624741):

```
======================================================================
FAIL: testtools.tests.test_testresult.TestTextTestResult.test_stopTestRun_current_time
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testtools/tests/test_testresult.py", line 1388, in test_stopTestRun_current_time
    DocTestMatches("... in 0.001s\n...", doctest.ELLIPSIS))
  File "testtools/testcase.py", line 435, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: Expected:
    ... in 0.001s
    ...
Got:
    <BLANKLINE>
    Ran 1 test in 0.000s
    OK
```

testtools/compat.py
rbtcollins: Yes, unless we alter setup.py to not require importing testtools. Which I'd like to do but is a separate conceptual change. Note that it was already broken before this patch - my other changes don't introduce the damage.

rbtcollins:
The test failure is not related to this code - note the 0.0000s vs 0.0001s.

testtools/content.py
rbtcollins: Done

testtools/content.py
rbtcollins: Its what the prior code did. The indent has shifted is all. Valid point but I'm meh either way today :)

testtools/content.py
rbtcollins: It's not. See the comment. I will do another patch to address setup.py and cleanup the existing (before this patch) and introduced by this patch conditionals. But it may require dropping features, not sure yet.

doc/for-framework-folk.rst
rbtcollins: Done

testtools/run.py
```diff
@@ -219,10 +224,18 @@ def _get_runner(self):
...
-                                         failfast=self.failfast,
-                                         buffer=self.buffer,
-                                         stdout=self.stdout)
+            try:
+                testRunner = self.testRunner(verbosity=self.verbosity,
+                                             failfast=self.failfast,
+                                             buffer=self.buffer,
+                                             stdout=self.stdout,
+                                             tb_locals=self.tb_locals)
+            except TypeError:
```
thomir: Ugh - this nested try / except is starting to look silly. I guess for now it's OK though. Let's hope we don't find a test runner that accepts 'tb_locals' but *not* 'verbosity :D

thomir:
LGTM - can't wait to have locals printed in testtools :D


Cheers,

rbtcollins:
I'm going to merge this now as it has Thomi's +1 and Jono had no API concerns in his feedback (and I addressed all the please-do-X stuff).

PR #131, opened 2015-03-10
Fix output of empty attachments.

rbtcollins:
Empty attachments to tests were triggering a file payload of None in the
``ExtendedToStreamDecorator`` code, which caused multiple copies of
attachments that had been output prior to the empty one.
(Robert Collins, #1378609)

Change-Id: I34a989546d57b245d4384b8c5b6d444e7ce0ac1b

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,14 @@ Changes and improvements to testtools_, grouped by release.
 NEXT
 ~~~~
 
+Improvements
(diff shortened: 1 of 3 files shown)


rbtcollins:
The test failure is the race on pypy3, not this change.

thomir:
LGTM

PR #133, opened 2015-03-11
Fix pypy3 tests by using ceil when reporting duration.

rbtcollins:
During reporting in ``TextTestResult`` now always uses ``ceil`` rather than
depending on the undefined rounding behaviour in string formatting.
(Robert Collins)

Change-Id: I72e11ccd1c41e9dbc65358aba5c1ffdc2d96eaf6

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,10 @@ Improvements
   This was a side effect of the fix to bug #941958, where we fixed a cosmetic
   error. (Robert Collins, #1430534)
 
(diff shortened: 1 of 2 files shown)


thomir:
I'm happy if travis is happy....

rbtcollins:
Travis is happy with the unittests. However sphinx upstream has released 1.3 and broken pypy3 and python3.2 https://pypi.python.org/pypi/Sphinx/1.3 :(

PR #137, opened 2015-04-02
Rewind the deferred test log before attaching it.

cjwatson:
Since the StringIO buffer is positioned at end-of-file after the
underlying test finishes, the twisted-log detail was previously always
empty.

### DIFF
--- a/testtools/deferredruntest.py
+++ b/testtools/deferredruntest.py
@@ -13,15 +13,10 @@ __all__ = [
     'SynchronousDeferredRunTest',
     ]
 
-import os
 import sys
 
 from testtools.compat import StringIO
-from testtools.content import (
-    Content,
-    text_content,
-    )
-from testtools.content_type import UTF8_TEXT
+from testtools.content import text_content
 from testtools.runtest import RunTest
 from testtools._spinner import (
     extract_result,
@@ -233,10 +228,8 @@ class AsynchronousDeferredRunTest(_DeferredRunTest):
             [error_observer.gotEvent, full_observer.emit],
             self._blocking_run_deferred, spinner)
 
-        full_log.seek(0, os.SEEK_SET)
-        log_content = full_log.readlines()
         self.case.addDetail(
-            'twisted-log', Content(UTF8_TEXT, lambda: log_content))
+            'twisted-log', text_content(full_log.getvalue()))
 
         logged_errors = error_observer.flushErrors()
         for logged_error in logged_errors:
--- a/testtools/deferredruntest.py
+++ b/testtools/deferredruntest.py
@@ -13,6 +13,7 @@ __all__ = [
     'SynchronousDeferredRunTest',
     ]
 
+import os
 import sys
 
 from testtools.compat import StringIO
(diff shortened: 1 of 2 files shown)


testtools/deferredruntest.py
```diff
@@ -232,8 +233,10 @@ def _run_core(self):
             [error_observer.gotEvent, full_observer.emit],
             self._blocking_run_deferred, spinner)
 
+        full_log.seek(0, os.SEEK_SET)
+        log_content = full_log.readlines()
         self.case.addDetail(
-            'twisted-log', Content(UTF8_TEXT, full_log.readlines))
+            'twisted-log', Content(UTF8_TEXT, lambda: log_content))
```
thomir: I don't see the need for the lambda here? Why not just keep this as it was:

 Content(UTF8_TEXT, full_log.readlines)

testtools/tests/test_deferredruntest.py
```diff
@@ -620,11 +622,17 @@ def test_something(self):
         result = self.make_result()
         runner.run(result)
         self.assertThat(
-            result._events,
+            [event[:2] for event in result._events],
```
thomir: I guess doing this will hide any point in the future where any of these events grow to be more than 2 items long. Not sure if that matters or not.

An alternative would be to get tricky with MatchesListwise I guess.

thomir:
Looks good, but I spotted a few things you might want to revisit. I'll let @lifeless have the final word.

testtools/deferredruntest.py
rbtcollins: If you're adding static content, there's a helper for that which will be more compact. text_content IIRC.

testtools/deferredruntest.py
```diff
@@ -232,8 +233,10 @@ def _run_core(self):
             [error_observer.gotEvent, full_observer.emit],
             self._blocking_run_deferred, spinner)
 
+        full_log.seek(0, os.SEEK_SET)
```
rbtcollins: Since we know full_log is a StringIO and since we're pulling the content out, calling full_log.getvalue() would be simpler and more efficient.

testtools/tests/test_deferredruntest.py
```diff
@@ -620,11 +622,17 @@ def test_something(self):
...
                 ('startTest', test),
-                ('addSuccess', test, {'twisted-log': text_content('')}),
+                ('addSuccess', test),
                 ('stopTest', test)]))
+        error = result._events[1][2]
+        self.assertThat(error, KeysEqual('twisted-log'))
+        self.assertThat(
+            error['twisted-log'].as_text(),
+            ContainsAll(
+                ['Traceback (most recent call last):', 'ZeroDivisionError']))
```
rbtcollins: I think this might be easier to read if we had a helper that materialised the content here, giving us a simple basic data type structure we can compare to - because there is no dynamic content in the test.

rbtcollins:
I've put a few thoughts inline. Overall - thanks for finding this bug, and the patch would appear to fix it. I'd like the fix to be made simpler, so as to avoid adding debt. The tests I am ok-as-is, though I've suggested a way to make it clearer and if you have the time to do that I'd appreciate it.

testtools/deferredruntest.py
cjwatson: If you set the Content instance's get_bytes to full_log.readlines, then it gets to iterate over it once and then will be at EOF if it tries to do it again.  This means that you get confusing results if, for example, you try to use an Equals matcher on it: it fails the match, then tries to iterate over the content again to construct the failure message and you just get the empty string.

But Robert's suggestions of using full_log.getvalue() and text_content() are better still, and make this moot.

thomir: Ahh, of course. Thanks
:-)

cjwatson:
The change to ADRT itself should be uncontroversial now.  Rather than taking advice exactly literally, I souped up the tests a bit with some custom matchers, and they feel a bit more accurate and clearer to me now; let me know what you think.

LiFeleSs:
uh?

thomir:
Sorry for the delay - LGTM

PR #140, opened 2015-05-06
Fix Bug #1452436 -- Code is now PEP 396 compliant

ddriddle:
`__version__` now returns a version string instead of a tuple per [PEP 396](https://www.python.org/dev/peps/pep-0396/). Added `__version_info__` which returns a version tuple. Fixes [Bug #1451989](https://bugs.launchpad.net/testtools/+bug/1452436).

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/140)

### DIFF
--- a/testtools/__init__.py
+++ b/testtools/__init__.py
@@ -123,5 +123,6 @@ else:
 # Otherwise it is major.minor.micro~$(revno).
 from pbr.version import VersionInfo
 _version = VersionInfo('testtools')
-__version__ = _version.semantic_version().version_tuple()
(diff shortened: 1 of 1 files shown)


jml:
Thanks! Discussion on the bug has it that we don't really want to do this, for backwards compatibility reasons. Does the pkg_resources / distutils stuff on the bug satisfy your use case?

rbtcollins:
Ping? If there's no more discussion my end feb, I'm going to close this to keep the list of PR's maintainable.

PR #141, opened 2015-05-11
BLD: Makefile: gh-pages, install[-{...}]

westurner:


rbtcollins:
Hi, this looks interesting. I presume its about setting up a git hub page for testtools? Nice.

A few thoughts/suggestions. Perhaps we could put the dev requirements into setup.cfg as an extras_require [pbr doesn't support extras yet].

 --find-links is deprecated in pip. Whats it for - whats the intent of that make target?

westurner:
> Hi, this looks interesting. I presume its about setting up a git hub page for testtools? Nice.

Thanks!

> A few thoughts/suggestions. Perhaps we could put the dev requirements into setup.cfg as an extras_require [pbr doesn't support extras yet].

That would be fine. I prefer the separate requirements.txt file solely because the pip install syntax is so varied w/ extras_require.

A more DRY solution: https://github.com/westurner/pyleset/blob/release/0.1.0/setup.py#L144

> --find-links is deprecated in pip. Whats it for - whats the intent of that make target?

Documentation, mostly. I'd have no problem removing them

rbtcollins:
Ouch no :) what you have is sufficient.

westurner:
Great. Do you need any changes from me?

rbtcollins:
Rather than asking you to tweak and tweak, I'm going through the changes locally, I'll rebase, tweak and then push up to master. Some of the targets make no sense to me (e.g. anything in the Makefile has to presume that you've already got the source, so install-from-github doesn't make sense).

rbtcollins:
[If there's a reason why it makes sense, we can add it in a future iteration]

rbtcollins:
Ok, so - I've pushed up a possible set of tweaks to https://github.com/rbtcollins/testtools/tree/westurner-patch-1 .

I've also manually build and pushed up the github page once - its now visible at http://testing-cabal.github.io/testtools/ so we can look at it and discuss it.

It seems to be identical in content to http://testtools.readthedocs.org/en/latest/ (which now I understand what this is doing in more detail is to be expected).

I think JML was thinking of something more modern-web-design aesthetic when he talked about github pages - after all, we already have the readthedocs site, which is 100% automated and we basically never need to think about.

I'm going to propose the dev extra and the docs I added to master as a separate patch, since I think they are independently useful.

westurner:
Either GitHub Pages or ReadTheDocs is great, thanks!

The extra build targets were for teaching purposes: if they're extraneous
then they are.

A link in the GitHub project description (and/or the URL field) would
probably be helpful as well.

rbtcollins:
Yes, we should advertise the rtd site better. I have no idea how discoverable it is today.

westurner:
> Some of the targets make no sense to me (e.g. anything in the Makefile has to presume that you've already got the source, so install-from-github doesn't make sense).

A place to store the commands to verify that a public project is working as expected by the documentation. Makefiles: Often ess than DRY, but maintainable.

> Yes, we should advertise the rtd site better. I have no idea how discoverable it is today.

To add a link to the docs to the GitHub Project (within the GitHub application):

- https://github.com/testing-cabal/testtools/
- Click 'Edit'
- The URL field accepts URLS (e.g. http://testtools.readthedocs.org/en/latest/ )
- The description field accepts text and URLs; and shows in the repository listing page (which may justify having two copies of the same link)


I've been adding links to projects in RST like this (cc'id here; thanks again; closing):

```rst

.. <sphinx>
.. index:: testtools
.. _testtools:

.. </sphinx>


testtools a
--------------
| Src: git https://github.com/testing-cabal/testtools
| PyPI: https://pypi.python.org/pypi/testtools
| Docs: https://testing-cabal.github.io/testtools/
| Docs: https://testtools.readthedocs.org/en/latest/
| Docs: https://testtools.readthedocs.org/en/latest/for-test-authors.html#assertions
| Docs: https://testtools.readthedocs.org/en/latest/for-test-authors.html#matchers
| Docs: https://testtools.readthedocs.org/en/latest/for-test-authors.html#running-your-tests


other heading
---------------------
* Link to :ref:`testtools` (italic link, locally dereferenced w/ next heading lookup)
* Link to `testtools a`_ (local link to heading of the same name)

#####

*****************************
other document (sphinx)
******************************
* Link to :ref:`testtools` (italic link, locally dereferenced w/ next heading lookup)
* Link to `other document (sphinx)`_
* Link to `<#other-document-sphinx>`__ (default slugified heading permalink)

.. note:: explicit ``.. _link:`` targets and ``:ref:`links <link>```
    are more maintainable (grep '^.. _te rm:'  && grep 'te rm')
    than absolute and relative URIs (with #heading-id-permalinks)

```

rbtcollins:
Added the docs link on github - lovely tip, thanks. Little confused on the other bit - do you mean in the README (which becomes the package description on PyPI) ?

PR #142, opened 2015-05-20
Ensure fixture cleanups are called even if setUp fails with useFixture

ZZelle:
This change ensures that a fixture cleanUp will be called even if its
setUp fails when called by useFixture.

Example:

  class Test(testtools.TestCase):
    def test(self):
      self.useFixture(fixture)

self.useFixture ensures that fixture.cleanUp will be called even if
fixture.setUp fails.

### DIFF
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -662,6 +662,7 @@ class TestCase(unittest.TestCase):
         :return: The fixture, after setting it up and scheduling a cleanup for
            it.
         """
+        self.addCleanup(fixture.cleanUp)
         try:
             fixture.setUp()
         except:
@@ -678,7 +679,6 @@ class TestCase(unittest.TestCase):
                 # encountered.
                 reraise(*exc_info)
         else:
-            self.addCleanup(fixture.cleanUp)
             self.addCleanup(
                 gather_details, fixture.getDetails(), self.getDetails())
             return fixture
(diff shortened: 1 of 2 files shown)


thomir:
Hi,

I guess the logic previously was that setUp should be an atomic operation, either completing fully or not changing the state of the fixture.

Your change looks sensible (travis seems to have failed in an interesting way, so I kicked off a second build). I'm +1 to merge, but will wait for @rbtcollins to comment on the new logic.

jml:
Thomi, this is indeed the logic: if `setUp` fails, it's responsible for cleaning up its mess. 

@rbtcollins had a reason for why that was more sensible than requiring `cleanUp` methods to behave sanely in the face of incomplete `setUp` calls, but I can't recall exactly what it was.

rbtcollins:
Thanks for your patience on this issue. I've written up an analysis[1] of the situation. I believe this patch here is incorrect: it doesn't solve the problem for users of fixtures, because part of the API is supporting 'with fixture:' - and the change here won't address that use case at all. I'm going to work up a patch to provide a more supportive setUp and so on within Fixtures. We'll want to change testtools to take advantage of the new exception that will be raised in these cases, but that can wait, most of the win will come from fixtures being updated to be more robust.

We will need to make cleanUp callable in the face of partially setup fixtures, but to support with: that has to be done from within setUp itself, unless we break symmetry between setUp/cleanUp and __enter__/__exit__ which would be very surprising IMO.

[1]: https://rbtcollins.wordpress.com/2015/06/22/revisiting-the-fixture-api-handling-leaky-resources/

rbtcollins:
https://github.com/testing-cabal/fixtures/pull/11 is my PR for fixtures. Assuming that goes in unaltered I'll prep an improvement patch for testtools to gather details from the exception.

PR #146, opened 2015-06-29
Fix the call to the parent setup.

elopio:
In the documentation example, the name of the class is wrong in the super call.


rbtcollins:
Thanks

PR #147, opened 2015-07-01
fixup tests with unittest2 1.1.0

rbtcollins:


thomir:
LGTM, assuming travis passes.

PR #149, opened 2015-08-19
Port twisted.deferredruntest to Twisted >= 15.1.0

cjwatson:
Twisted 15.1.0 removes the compatibility import of _LogObserver in
twisted.trial.unittest.  This is unfortunate for us, but it's what we
get for using an internal interface.  It at least still exists in
twisted.trial._synctest, so we can get it from there.

Twisted 15.2.0 adds the new twisted.logger framework, which requires a
slight adjustment to run_with_log_observers.  There's no longer a
supported interface to get hold of all log observers, but since we're
already using an internal interface (see above), what's one more?

This passes "make check" with the current release, Twisted 15.3.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,21 +18,21 @@ python:
 matrix:
   include:
     - python: "2.6"
-      env:
-        - TWISTED_REQ="Twisted==13.0.0"
-        - TWISTED_REQ="Twisted==15.2.0"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "2.6"
+      env: TWISTED_REQ="Twisted==15.2.0"
     - python: "2.7"
-      env:
-        - TWISTED_REQ="Twisted==13.0.0"
-        - TWISTED_REQ="Twisted==15.2.0"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "2.7"
+      env: TWISTED_REQ="Twisted==15.2.0"
     - python: "3.2"
       env:
         - JINJA_REQ="jinja2<2.7, Pygments<2.0"
         - SPHINX="<1.3"
     - python: "pypy"
-      env:
-        - TWISTED_REQ="Twisted==13.0.0"
-        - TWISTED_REQ="Twisted==15.2.0"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "pypy"
+      env: TWISTED_REQ="Twisted==15.2.0"
     - python: "pypy3"
       env: SPHINX="<1.3"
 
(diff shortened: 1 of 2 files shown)


rbtcollins:
Does it still work with older supported Twisted releases?

rbtcollins:
(I mean, I see the conditional code, but we don't strictly know). I think to merge this we probably want a new entry in the test matrix to run old twisted

cjwatson:
It worked for me back to at least 14.0.0, but I take your point.  How would I go about extending the test matrix?  Is that even something in my power to do?

cjwatson:
OK, I worked it out.  Could you have a look at this?

.travis.yml
```diff
@@ -9,21 +9,35 @@ python:
...
 # See also:
 # http://stackoverflow.com/questions/18252804/syntax-error-in-jinja-2-library
+#
+# Twisted tests currently only work on Python 2.
 matrix:
   include:
+    - python: "2.6"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "2.6"
+      env: TWISTED_REQ="Twisted==15.2.0"
```
rbtcollins: Ok so this one should be uncapped I think. Just 'Twisted'.

.travis.yml
```diff
@@ -9,21 +9,35 @@ python:
...
+    - python: "2.7"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "2.7"
+      env: TWISTED_REQ="Twisted==15.2.0"
```
rbtcollins: ditto

.travis.yml
```diff
@@ -9,21 +9,35 @@ python:
...
     - python: "3.2"
       env:
         - JINJA_REQ="jinja2<2.7, Pygments<2.0"
         - SPHINX="<1.3"
+    - python: "pypy"
+      env: TWISTED_REQ="Twisted==13.0.0"
+    - python: "pypy"
+      env: TWISTED_REQ="Twisted==15.2.0"
```
rbtcollins: ditto.

rbtcollins:
Do we need a test of the new global log observer functionality? Perhaps we should also file a ticket in twisted on the privates we're having to use.

cjwatson:
My feeling is that the fact that a slew of tests were failing without this change is probably sufficient - that is, the current deferredruntest tests already covered the basic function of running with different log observers, it's just that that needs different implementation details with current Twisted.

rbtcollins:
I have filed https://twistedmatrix.com/trac/ticket/8048 about the API being non-public.

rbtcollins:
Merged, thanks!

PR #151, opened 2015-09-10
Add show-import-errors parameter

aplanas:
Running `subunit.run discovery --list` hide the reason behing errors
during the test import.  Is expected that, like subunit, other
TestRunners will exist (sys.exit(2)) in similar sitations.

This patch provide a new CLI parameter `--show-import-errors` to
force the report of the different import errors.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/151)

### DIFF
--- a/testtools/run.py
+++ b/testtools/run.py
@@ -158,6 +158,7 @@ class TestProgram(unittest.TestProgram):
         self.defaultTest = defaultTest
         # XXX: Local edit (see http://bugs.python.org/issue22860)
         self.listtests = False
+        self.showerrors = False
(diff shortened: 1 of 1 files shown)


rbtcollins:
Sorry, I don't have the context for this. Is there a bug to refer to ? It doesn't make sense to me, since subunit.run's output is programmatically consumed, and I'd expect the consumer to show users the errors.

rbtcollins:
Ping @aplanas

rbtcollins:
Closing, month with no reply.

PR #154, opened 2015-10-05
Update hacking documentation.

thomir:


doc/hacking.rst
```diff
@@ -39,6 +39,18 @@ Assignment above) needs to be licensed under the `MIT license`_ that testtools
 uses, so that testtools can ship it.
 
 
+Installing
```
rbtcollins: I don't think installing makes sense *here* in the docs: this is docs on hacking, and the 'testing' section below covers this.

Secondly, for users, I think we want to say 'pip install testtools' - anyone installing from source should be coming from a position of knowledge after all...

thomir:
@rbtcollins Good point.

I've updated the heading to 'Building', de-duplicated the pip installation instruction. This means that the 'Testing' section is now solely about running and writing tests, and the 'Building' section now contains all the pip etc. content.

Thoughts?

rbtcollins:
Squashed and rebased and landed.

thomir:
Thanks @rbtcollins  - I would have squashed the commits myself had you approved.

rbtcollins:
That would have been fine too; I just didn't see any point in a hand-off once it was good :)

PR #163, opened 2015-11-13
runtest: add cProfile support

jd:
This patch allows to profile unit tests using cProfile. If the
environment variable TESTTOOLS_CPROFILE is set, a `cprof` file is
generated with the full name of the test, containing the profiling
information.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/163)

### DIFF
--- a/testtools/runtest.py
+++ b/testtools/runtest.py
@@ -7,7 +7,9 @@ __all__ = [
     'RunTest',
     ]
 
+import cProfile
 import sys
+import os
 
 from testtools.testresult import ExtendedToOriginalDecorator
 
@@ -64,6 +66,11 @@ class RunTest(object):
         self.exception_caught = object()
         self._exceptions = []
         self.last_resort = last_resort or (lambda case, result, exc: None)
+        self._cprof_dir = os.environ.get("TESTTOOLS_CPROFILE")
+        if os.environ.get("TESTTOOLS_CPROFILE"):
+            self._pr = cProfile.Profile()
+        else:
+            self._pr = None
 
     def run(self, result=None):
         """Run self.case reporting activity to result.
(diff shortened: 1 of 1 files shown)


jml:
@jd Thanks for the patch. 

What were your reasons for implementing it in `RunTest`? 

I ask because I think it should be something separate. Perhaps either as something that wraps an existing `RunTest` (although I've just tried this recently, and it's a lot harder & fiddlier than it should be), or as a `Fixture`.

testtools/runtest.py
```diff
@@ -76,9 +83,17 @@ def run(self, result=None):
...
+            self._pr.enable()
         try:
             return self._run_one(actual_result)
         finally:
+            if self._pr:
+                self._pr.disable()
+                self._pr.dump_stats("%s.%s.%s.cprof"
+                                    % (self.case.__class__.__module__,
+                                       self.case.__class__.__name__,
+                                       self.case._get_test_method().__name__))
```
jml: This should be attached as a detail. AIUI, right now it just prints to stdout.

jd:
Hi @jml! This is a very naive implementation, so any suggestion welcome. I'd love to have that automagically available to any project that uses testtools, so that's why I implemented it this way.

I don't think using a `Fixture` would allow that?

rbtcollins:
@jd thanks for the patch. I think using an environment variable to activate this is a bi of a rough UI. I'd rather see testtools.run grow options to enable / disable profiling.

jono's comments about code structure don't really alter wether this is builtin or not, but will help keep concerns that should be separate separate.

testtools/runtest.py
jd: @jml no, it prints it out to a file. But attaching to a result sounds like a good idea.

jml:
@jd Thanks.

@rbtcollins is right in that 

Putting together what @rbtcollins and I suggested:
* Please move the logic to a `Fixture`
* Have that fixture return the data as a detail
*

jml:
I keep hitting the wrong button!

rbtcollins:
I'm not clear; this seems like we've said 'you should do this externally'? If so, may we close the PR?

jml:
I guess I'd be OK with merging in a fixture that does what this does, but probably it would be better for that to be implemented as part of fixtures itself, or its own library. So, yeah, close for now.

PR #165, opened 2015-11-16
Allow tests to be run more than once

jml:
Our `setUp` and `tearDown` protection would prevent a single test from being run more than once. This makes a `RunTest`-based solution to [#1515933 - Support for retrying intermittently failing tests](https://bugs.launchpad.net/testtools/+bug/1515933) rather difficult.

This patch makes it possible to re-run tests, failed or otherwise. AFAICT, it doesn't change any of our public APIs.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/165)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,7 +10,13 @@ Changes
 -------
 
 * Add a new test dependency of testscenarios. (Robert Collins)
-  
+
+* ``addCleanup`` can now only be called during a test run.
+  (Jonathan Lange)
+
+* ``TestCase`` objects can now be run twice. All internal state is reset
+  between runs.  (Jonathan Lange)
+
 1.8.1
 ~~~~~
 
--- a/NEWS
+++ b/NEWS
@@ -29,7 +29,7 @@ Changes
 
 * Add a new test dependency of testscenarios. (Robert Collins)
 
-* ``addCleanup`` can now only be called during a test run.
+* ``addCleanup`` can now only be called within a test run.
   (Jonathan Lange)
 
 * ``TestCase`` objects can now be run twice. All internal state is reset
--- a/testtools/tests/helpers.py
+++ b/testtools/tests/helpers.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2008-2012 testtools developers. See LICENSE for details.
+# Copyright (c) 2008-2016 testtools developers. See LICENSE for details.
 
 """Helpers for tests."""
 
@@ -29,6 +29,7 @@ try:
     raise Exception
 except Exception:
     an_exc_info = sys.exc_info()
+
 
 # Deprecated: This classes attributes are somewhat non deterministic which
 # leads to hard to predict tests (because Python upstream are changing things.
@@ -153,9 +154,14 @@ class AsText(AfterPreprocessing):
             lambda log: log.as_text(), matcher, annotate=annotate)
 
 
-def throw(function, *args, **kwargs):
-    """Call ``function`` with arguments and raise the result.
+def raise_(exception):
+    """Raise ``exception``.
 
-    Use this when you want to raise within an expression.
+    Useful for raising exceptions when it is inconvenient to use a statement
+    (e.g. in a lambda).
+
+    :param Exception exception: An exception to raise.
+    :raises: Whatever exception is
+
     """
-    raise function(*args, **kwargs)
+    raise exception
--- a/testtools/tests/helpers.py
+++ b/testtools/tests/helpers.py
@@ -151,3 +151,11 @@ class AsText(AfterPreprocessing):
     def __init__(self, matcher, annotate=True):
         super(AsText, self).__init__(
             lambda log: log.as_text(), matcher, annotate=annotate)
+
+
+def throw(function, *args, **kwargs):
+    """Call ``function`` with arguments and raise the result.
+
+    Use this when you want to raise within an expression.
+    """
+    raise function(*args, **kwargs)
--- a/testtools/tests/helpers.py
+++ b/testtools/tests/helpers.py
@@ -12,6 +12,12 @@ from extras import safe_hasattr
 
 from testtools import TestResult
 from testtools.content import StackLinesContent
+from testtools.matchers import (
+    AfterPreprocessing,
+    Equals,
+    MatchesDict,
+    MatchesListwise,
+)
 from testtools import runtest
 
 
@@ -106,3 +112,42 @@ class FullStackRunTest(runtest.RunTest):
         return run_with_stack_hidden(
             False,
             super(FullStackRunTest, self)._run_user, fn, *args, **kwargs)
+
+
+class MatchesEvents(object):
+    """Match a list of test result events.
+
+    Specify events as a data structure.  Ordinary Python objects within this
+    structure will be compared exactly, but you can also use matchers at any
+    point.
+    """
+
+    def __init__(self, *expected):
+        self._expected = expected
+
+    def _make_matcher(self, obj):
+        # This isn't very safe for general use, but is good enough to make
+        # some tests in this module more readable.
+        if hasattr(obj, 'match'):
+            return obj
+        elif isinstance(obj, tuple) or isinstance(obj, list):
+            return MatchesListwise(
+                [self._make_matcher(item) for item in obj])
+        elif isinstance(obj, dict):
+            return MatchesDict(dict(
+                (key, self._make_matcher(value))
+                for key, value in obj.items()))
+        else:
+            return Equals(obj)
+
+    def match(self, observed):
+        matcher = self._make_matcher(self._expected)
+        return matcher.match(observed)
+
+
+class AsText(AfterPreprocessing):
+    """Match the text of a Content instance."""
+
+    def __init__(self, matcher, annotate=True):
+        super(AsText, self).__init__(
+            lambda log: log.as_text(), matcher, annotate=annotate)
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -201,7 +201,6 @@ class TestCase(unittest.TestCase):
         """
         runTest = kwargs.pop('runTest', None)
         super(TestCase, self).__init__(*args, **kwargs)
-        self._cleanups = []
         self._reset()
         test_method = self._get_test_method()
         if runTest is None:
@@ -224,6 +223,7 @@ class TestCase(unittest.TestCase):
 
     def _reset(self):
         """Reset the test case as if it had never been run."""
+        self._cleanups = []
         self._unique_id_gen = itertools.count(1)
         # Generators to ensure unique traceback ids.  Maps traceback label to
         # iterators.
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -201,11 +201,13 @@ class TestCase(unittest.TestCase):
         """
         runTest = kwargs.pop('runTest', None)
         super(TestCase, self).__init__(*args, **kwargs)
+        # XXX: jml: Before resubmitting, move these into _reset. We don't
+        # (yet!) have tests driving the change, but it does seem like the more
+        # correct thing to do.
         self._cleanups = []
         self._unique_id_gen = itertools.count(1)
         # Generators to ensure unique traceback ids.  Maps traceback label to
         # iterators.
-        self._traceback_id_gens = {}
         self._reset()
         test_method = self._get_test_method()
         if runTest is None:
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -639,6 +639,8 @@ class TestCase(unittest.TestCase):
                 "super(%s, self).tearDown() from your tearDown()."
                 % (sys.modules[self.__class__.__module__].__file__,
                    self.__class__.__name__))
+        self.__setup_called = False
+        self.__teardown_called = False
         return ret
 
     def _get_test_method(self):
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -630,18 +630,19 @@ class TestCase(unittest.TestCase):
         :raises ValueError: If the base class tearDown is not called, a
             ValueError is raised.
         """
-        ret = self.tearDown()
-        if not self.__teardown_called:
-            raise ValueError(
-                "In File: %s\n"
-                "TestCase.tearDown was not called. Have you upcalled all the "
-                "way up the hierarchy from your tearDown? e.g. Call "
-                "super(%s, self).tearDown() from your tearDown()."
-                % (sys.modules[self.__class__.__module__].__file__,
-                   self.__class__.__name__))
-        self.__setup_called = False
-        self.__teardown_called = False
-        return ret
+        try:
+            return self.tearDown()
+        finally:
+            if not self.__teardown_called:
+                raise ValueError(
+                    "In File: %s\n"
+                    "TestCase.tearDown was not called. Have you upcalled all the "
+                    "way up the hierarchy from your tearDown? e.g. Call "
+                    "super(%s, self).tearDown() from your tearDown()."
+                    % (sys.modules[self.__class__.__module__].__file__,
+                       self.__class__.__name__))
+            self.__setup_called = False
+            self.__teardown_called = False
 
     def _get_test_method(self):
         method_name = getattr(self, '_testMethodName')
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -201,13 +201,8 @@ class TestCase(unittest.TestCase):
         """
         runTest = kwargs.pop('runTest', None)
         super(TestCase, self).__init__(*args, **kwargs)
-        # XXX: jml: Before resubmitting, move these into _reset. We don't
-        # (yet!) have tests driving the change, but it does seem like the more
-        # correct thing to do.
         self._cleanups = []
         self._unique_id_gen = itertools.count(1)
-        # Generators to ensure unique traceback ids.  Maps traceback label to
-        # iterators.
         self._reset()
         test_method = self._get_test_method()
         if runTest is None:
@@ -230,6 +225,8 @@ class TestCase(unittest.TestCase):
 
     def _reset(self):
         """Reset the test case as if it had never been run."""
+        # Generators to ensure unique traceback ids.  Maps traceback label to
+        # iterators.
         self._traceback_id_gens = {}
         self.__setup_called = False
         self.__teardown_called = False
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -636,19 +636,16 @@ class TestCase(unittest.TestCase):
         :raises ValueError: If the base class tearDown is not called, a
             ValueError is raised.
         """
-        try:
-            return self.tearDown()
-        finally:
-            if not self.__teardown_called:
-                raise ValueError(
-                    "In File: %s\n"
-                    "TestCase.tearDown was not called. Have you upcalled all the "
-                    "way up the hierarchy from your tearDown? e.g. Call "
-                    "super(%s, self).tearDown() from your tearDown()."
-                    % (sys.modules[self.__class__.__module__].__file__,
-                       self.__class__.__name__))
-            self.__setup_called = False
-            self.__teardown_called = False
+        ret = self.tearDown()
+        if not self.__teardown_called:
+            raise ValueError(
+                "In File: %s\n"
+                "TestCase.tearDown was not called. Have you upcalled all the "
+                "way up the hierarchy from your tearDown? e.g. Call "
+                "super(%s, self).tearDown() from your tearDown()."
+                % (sys.modules[self.__class__.__module__].__file__,
+                   self.__class__.__name__))
+        return ret
 
     def _get_test_method(self):
         method_name = getattr(self, '_testMethodName')
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -206,11 +206,7 @@ class TestCase(unittest.TestCase):
         # Generators to ensure unique traceback ids.  Maps traceback label to
         # iterators.
         self._traceback_id_gens = {}
-        self.__setup_called = False
-        self.__teardown_called = False
-        # __details is lazy-initialized so that a constructed-but-not-run
-        # TestCase is safe to use with clone_test_with_new_id.
-        self.__details = None
+        self._reset()
         test_method = self._get_test_method()
         if runTest is None:
             runTest = getattr(
@@ -229,6 +225,15 @@ class TestCase(unittest.TestCase):
             (_UnexpectedSuccess, self._report_unexpected_success),
             (Exception, self._report_error),
             ]
+
+    def _reset(self):
+        """Reset the test case as if it had never been run."""
+        self._traceback_id_gens = {}
+        self.__setup_called = False
+        self.__teardown_called = False
+        # __details is lazy-initialized so that a constructed-but-not-run
+        # TestCase is safe to use with clone_test_with_new_id.
+        self.__details = None
 
     def __eq__(self, other):
         eq = getattr(unittest.TestCase, '__eq__', None)
@@ -596,6 +601,7 @@ class TestCase(unittest.TestCase):
         result.addUnexpectedSuccess(self, details=self.getDetails())
 
     def run(self, result=None):
+        self._reset()
         try:
             run_test = self.__RunTest(
                 self, self.exception_handlers, last_resort=self._report_error)
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -202,7 +202,6 @@ class TestCase(unittest.TestCase):
         runTest = kwargs.pop('runTest', None)
         super(TestCase, self).__init__(*args, **kwargs)
         self._cleanups = []
-        self._unique_id_gen = itertools.count(1)
         self._reset()
         test_method = self._get_test_method()
         if runTest is None:
@@ -225,6 +224,7 @@ class TestCase(unittest.TestCase):
 
     def _reset(self):
         """Reset the test case as if it had never been run."""
+        self._unique_id_gen = itertools.count(1)
         # Generators to ensure unique traceback ids.  Maps traceback label to
         # iterators.
         self._traceback_id_gens = {}
--- a/testtools/tests/samplecases.py
+++ b/testtools/tests/samplecases.py
@@ -1,0 +1,31 @@
+# Copyright (c) 2015 testtools developers. See LICENSE for details.
+
+"""A collection of sample TestCases.
+
+These are primarily of use in testing the test framework.
+"""
+
+from testtools import TestCase
+
+
+class _NormalTest(TestCase):
+
+    def test_success(self):
+        pass
+
+    def test_error(self):
+        1/0
+
+    def test_failure(self):
+        self.fail('arbitrary failure')
+
+
+"""
+A list that can be used with testscenarios to test every kind of sample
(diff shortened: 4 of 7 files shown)


rbtcollins:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[testtools/testcase.py, line 643 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFEwDNTFSamFBpGSa)** ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/testcase.py#L643)):</sup>
Do we want this to work even in the case of tearDown failing? If so it should be behind a finally. I'm fine as-is, just asking ;)


---

<sup>**[testtools/tests/test_runtest.py, line 86 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFZligyYiKlRsZmZY)** ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_runtest.py#L86)):</sup>
I'm a little concerned about this change.  This test is testing _run_prepared_result, a private API. By changing this to check tearDown, its making more assumptions about _run_teardown. I see why you need to change it: I suggest overriding _run_teardown in the test class instead, since that is the contract RunTest interacts with. It doesn't itself know about tearDown.


---

<sup>**[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L)** ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):</sup>
I think you need a scenario in here: they should be able to be run twice even if the following things happen, no?:

{stage} x {final state}

where stage in {setup, body, teardown, cleanups}
and final state in {pass, fail, error, skip, xfail}

(Obviously user code could completely break things like e.g. an external DB or something, but our framework should handle these cases).


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)

jml:
* Any external project that uses the attributes with double underscore knows to expect this sort of trouble
* Enabling new functionality of running a test twice won't cause another project to break, I don't think

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

_[testtools/testcase.py, line 643 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFEwDNTFSamFBpGSa:-K3JH8KzEr_zUxq9J2it:1353794417) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/testcase.py#L643)):_
Yes we do. This is in support of retrying intermittently failing tests. Such tests are just as likely to fail in teardown due to complicated fixtures as they are during the tests themselves. 

I've fixed this. Writing the test for `tearDown` requires more care than the regular test.

---

_[testtools/tests/test_runtest.py, line 86 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFZligyYiKlRsZmZY:-K3L-LCFCXHunnzGhEQO:899233299) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_runtest.py#L86)):_
Good point! Fixed. (How would you feel about having explicit interfaces/abcs in testtools? I think it's worth giving these contracts names.)

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3JHgyhOt2zFa_UD71u:1652279505) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Do we already have something for that?

I've been thinking for a while that we do need something like this for a lot of purposes (perhaps powered by Hypothesis). It could even go hand-in-hand with some kind of explicit documentation of interfaces. 

Anyway, if we already have something, I'm happy to make this change. If not, I'd rather do it in a follow-up patch, but maybe can be persuaded to cobble something together. (I actually wrote a couple of extra tests for different paths but none of them triggered failures so I dropped them). 

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-:-K3O_8tTr0em7aTKJ_V9:132994092)_

jml:
Thanks for the reviews!

rbtcollins:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3OcMjZd_IUdAM0vQDn:-1863519732) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
I think testscenarios is a good fit here, since its a simple product of cases.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3QKV2aIblFhF-9R2Zs:-247854476) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
OK, that means we probably don't already have something, since we don't currently depend on testscenarios or use it in any way. Leaving this as future work. 

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:

---

_[testtools/tests/test_runtest.py, line 86 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFZligyYiKlRsZmZY:-K3QqJCJAw81r2Z0Df_f:-1252421507) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_runtest.py#L86)):_
If they're lightweight I'm happy with it. Also PEP-484 annotations.

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3QqaA_N2I1ozgyDNte:530082582) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Well we depend on fixtures for testing, I see no reason we shouldn't depend on testscenarios too.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:

---

_[testtools/tests/test_runtest.py, line 86 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HFZligyYiKlRsZmZY:-K3QuKIgchGelGLB94Ya:1794731161) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_runtest.py#L86)):_
Great!

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3Qs83SW4q18IQj6MTy:580245572) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Happy w/ testscenarios—in a later patch.

My issue hasn't been about which technology we should use for doing this. It's that I don't want to predicate this patch on introducing a whole new class of testing strategy into the code-base, _even though I think it'd be a great idea_.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3R09JRukYd5eG7wQoT:-641374308) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Up to you on how you do it. Its going to be a nasty for loop with manual diagnostics, or 5 lines of extra diff (which I can write up for you if you like).

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3R3hp-alJTgYlk580s:974726543) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
https://github.com/testing-cabal/testtools/pull/166

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3R9ZglijbSw-Nl1xXC:35818674) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Thanks. I'll go ahead and write those tests at some point in the future. 

Why you think out of all the code that we have that would benefit from scenarios (I think there's lots), that this patch needs to be blocked on it?

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3RJj7GnnmWrWBNQq8v:-438845037) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
I don't think it has to be blocked on testscenarios. I think it has to be blocked on appropriate coverage for your declared goal,or we'll be fighting little brushfires as folk find holes in the next few releases.

I think doing that coverage via scenarios is less total effort than doing it by hand, with clearer results, so was suggesting you take what I perceive to be the path of least resistance in achieving your declared go.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3Th_e9gkDkf-WoMYg9:959728906) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Ah OK. We disagree about appropriate coverage.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3TjcNWq9Oa7_gykp_3:-737652602) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Ah. So here's my reasoning. I found a genuine bug by code inspection, thus I know the tests you have were insufficient. I'm not confident there are no other bugs lurking...

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K3Tsp2XoTd44LgIq9up:93024603) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):_
Oops, forgot to push.

Before I even submitted this patch, I actually wrote two more tests almost exactly like `test_runTwice`, including one specifically written for the case of erroring in `tearDown` that looked a lot like `test_runTwice`. Neither failed, so I deleted them.

A naive parametrized version of `test_runTwice` would have found not found that bug.

Anyway, I'll write the tests. 

These tests *might* drive a design change though. `getDetails()` accumulates over the course of many runs, and for truly repeatable runs it has to be cleared out. For compatibility with other tools, it would be good if that was done as part of the `run()` method. 

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

rbtcollins:
Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

_[testtools/testcase.py, line 644 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3W95cbfkXyyXJJMz63:-K3W95cbfkXyyXJJMz64:296782651) ([raw file](https://github.com/testing-cabal/testtools/blob/df8b8c0c9e3de2c8cff3605e4b5f440e5a368d25/testtools/testcase.py#L644)):_
```
class BreakSetup(TestCase):
    called = False

    def setUp(self):
        if self.called:
            return
        super(BreakSetup, self).setUp()

    def test_nil(self):
        pass

    def tearDown(self):
        if self.called:
            super(BreakSetup, self).tearDown()
        self.__class__.called = True

```
This simulates a test that fails to upcall in setup if some global state is borked, and fails to call tearDown at all. I believe if run twice with your patch the first time it will trigger the ValueError, and then the second time when it does not upcall in setUp, it won't trigger an error even though its now buggy. I've seen this sort of thing dealing with e.g. locked databases, so I  think its important to be correct here. I suspect its in the pathological space where hypothesis would be a better fit than entire-space-checking, but this sort of stateful interaction is probably something we need coverage on too.

tl;dr, the finally needs to guard the __setup_called and __teardown_called resetting, only. Or move it to the start of run() or something.

---


_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)_

jml:
Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[testtools/testcase.py, line 644 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3W95cbfkXyyXJJMz63:-K4N0uyJaC4PHP0gxXer:2117623833) ([raw file](https://github.com/testing-cabal/testtools/blob/df8b8c0c9e3de2c8cff3605e4b5f440e5a368d25/testtools/testcase.py#L644)):*
* Added a scenario with this case
* Moved cleanup to the start of run

---

*[testtools/tests/test_testcase.py, line 1269 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K3HGgB_HlLSrY_2-I5L:-K4N11FtjHlyaG0-3Dfv:-1177011245) ([raw file](https://github.com/testing-cabal/testtools/blob/cc7f66594f6f04dbaf6d207d91cf7da378618070/testtools/tests/test_testcase.py#L1269)):*
`getDetails()` clearing now happens in the reset. 

Parametrized details added.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)*

jml:
Build failure is unrelated to this branch and is fixed by #170. Please take a look at this branch.

---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-:-K4YZ2kx76rlcEhalqJL:1539461983)*

rbtcollins:
I think this is a big improvement. I would really love to see a matrix on the deterministic tests - something as simple as:

```
stages = [('setup', {'stage':'setup'}),('body', {'stage':'body'}),('teardown', {'stage':'teardown'}),('cleanups', {'stage':'cleanups'})]
states = [('pass', {'state':'pass'}),('fail', {'state':'fail'}),('error', {'state':'error'}),('skip', {'state':'skip'}),('xfail', {'state':'xfail'}),]

class Exhaustive(TestCase):

    class Harness(TestCase):

        def __init__(self, testmethodName, stage, state):
            self.stage = stage
            self state = state
            super(Harness, self).__init__(testmethodName)
        def outcome(self):
            if self.state == 'pass':
                return
            if self.state == 'fail':
                self.fail('failed')
            if self.state == 'error':
                1/0
            if self.state == 'skip':
                self.skip('skipped')
            if self.state == 'xfail':
                self.expectFailure("expected", self.assertNotEqual, 1, 0)
        def test_case(self):
            if self.stage != 'body':
                return
            self.outcome()
        def setUp(self):
            super(Harness, self).setUp()
            if self.stage == 'cleanups':
                self.addCleanup(self.outcome)
            if self.stage != 'setup':
                return
            self.outcome()
        def tearDown(self):
            super(Harness, self).tearDown()
            if self.stage != 'teardown':
                return
            self.outcome()

    scenarios = multiply_scenarios(stages, states)

    def test_scenarios(self):
        case = Harness('test_case', self.stage, self.state)
        ... run it and cross check events

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[testtools/testcase.py, line 205 \[r3\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4esHtkdspX6r5Dl5R1:-K4esHtkdspX6r5Dl5R2:776674581) ([raw file](https://github.com/testing-cabal/testtools/blob/c88d030f59fe707a889c4d5e7f6c8da9e3f97623/testtools/testcase.py#L205)):*
These three things (2 above, one below) should go in _reset too I think. The id_gens one is already there, redundantly.

---

*[testtools/tests/samplecases.py, line 45 \[r3\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4esoeg_tEzvnwiT91d:-K4esoeg_tEzvnwiT91e:1153736470) ([raw file](https://github.com/testing-cabal/testtools/blob/c88d030f59fe707a889c4d5e7f6c8da9e3f97623/testtools/tests/samplecases.py#L45)):*
"fails to call ``tearDown`` when the global state breaks but calls it after that.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-:-K4etR_ezfEPGaYrpFRK:1911037791)*
<!-- Sent from Reviewable.io -->

**Comment by jml** [comment]
Ah, OK, you want the full monty.

I've used your approach as a starting point (thanks so much!) and have gone for a more functional approach. 

Also (unless I'm misreading), the approach outlined here only every has stuff going on in one stage per test. The patch I'm about to push (I think) does all combinations of all stages. This more than doubles the number of tests in our suite!

---

Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[testtools/testcase.py, line 205 \[r3\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4esHtkdspX6r5Dl5R1:-K4iEuiFlnARPRCYTt83:-822364585) ([raw file](https://github.com/testing-cabal/testtools/blob/c88d030f59fe707a889c4d5e7f6c8da9e3f97623/testtools/testcase.py#L205)):*
Moved `_unique_id_gen` and `_traceback_id_gens`. It's definitely cleaner. FWIW I held off before because I couldn't find a test to drive the change.

Moving `self._cleanups = []` to `_reset` causes `TestAddCleanup` to fail. Currently we allow users to call `addCleanup` on a constructed `TestCase` before running and for those cleanups to be applied in the run. Moving it to `_reset` causes them to be cleared before the run. 

I'm mildly averse to this change, because it breaks compatibility.

---

*[testtools/tests/samplecases.py, line 45 \[r3\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4esoeg_tEzvnwiT91d:-K4ghohWLzAk9q8G8Z7D:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/c88d030f59fe707a889c4d5e7f6c8da9e3f97623/testtools/tests/samplecases.py#L45)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-:-K4iDeggxsyW6TDF5uRP:-1262264844)*

**Comment by rbtcollins** [comment]
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[testtools/testcase.py, line 204 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4t73jDBEM-r-wuYsaT:-K4t73jElHByAZUMCRpg:498194212) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/testcase.py#L204)):*
Ok so - I see your argument that we used to allow injecting cleanups before run()... but we never intended to support that, did we? Its clearly a skew from the contract of 'run runs the test, and tests are hermetic [module setupClass (OMG) and setupModule (crying into my wine). I think making this more hermetic is strictly good, and no less a break that our requiring that setUp be called.

e.g. our tests were using a naughty shortcut, naughty tests.

---

*[testtools/tests/samplecases.py, line 50 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tA784s8GGGffzmVi2:-K4tA784s8GGGffzmVi3:1519881582) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L50)):*
All of the stages $what?

---

*[testtools/tests/samplecases.py, line 84 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tAOvSon28ME8QY5gl:-K4tAOvSon28ME8QY5gm:1706606190) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L84)):*
Perhaps this should just be _do_nothing = _success ?

---

*[testtools/tests/samplecases.py, line 104 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tAWGcFSs8hzshTwYJ:-K4tAWGcFSs8hzshTwYK:98699437) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L104)):*
Or alternative change this to use _do_nothing - either way this is missing ", case" near the right hand end of the line...  (and if tests are passing, there is something wrong :)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)*

**Comment by jml** [comment]
Review status: 0 of 4 files reviewed at latest revision, 5 unresolved discussions.

---

*[testtools/testcase.py, line 204 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4t73jDBEM-r-wuYsaT:-K4vwr4_GcbE_y_DRPEf:-1644950041) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/testcase.py#L204)):*
It's not just _our_ tests. Nevertheless, changed. Significant updates to `TestAddCleanup` and `TestPatchSupport`. 

---

*[testtools/tests/samplecases.py, line 50 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tA784s8GGGffzmVi2:-K4wF5tud2L2x2inwuz_:594663406) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L50)):*
I've updated the docstring, and added one to the constructor as well.

---

*[testtools/tests/samplecases.py, line 84 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tAOvSon28ME8QY5gl:-K4wF2sStkykMSs0VpxK:901802839) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L84)):*
Changed to `_success = _do_nothing`.

---

*[testtools/tests/samplecases.py, line 104 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/165#-K4tAWGcFSs8hzshTwYJ:-K4wEb6JnWkrvCWPR4yC:267350997) ([raw file](https://github.com/testing-cabal/testtools/blob/db7e382f815d218b529b97e0118a003375d46ba9/testtools/tests/samplecases.py#L104)):*
Kept `_success` because I think that more clearly communicates intent.

Added `case` parameter—good catch.

The tests _were_ passing, but that's not necessarily an indication that something was wrong. A test that makes an invalid call to `expectFailure` is treated as erroring. By getting `_unexpected_success` wrong, we were effectively exercising the "error" case twice.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/165)*

**Comment by jml** [comment]
AFAICT, this only needs a rubber stamp to confirm that the last round of changes satisfy the last round of review.

File: NEWS [hunk 1]
```diff
@@ -29,6 +29,14 @@ Changes
 
 * Add a new test dependency of testscenarios. (Robert Collins)
 
+* ``addCleanup`` can now only be called during a test run.
```
rbtcollins: nit - perhaps 
'called within test.run()' I realise thats technically ambiguous for twisted with its non-generator async idiom, but I think its still clearer?

testtools/tests/helpers.py
```diff
@@ -106,3 +112,50 @@ def _run_user(self, fn, *args, **kwargs):
...
+
+class AsText(AfterPreprocessing):
+    """Match the text of a Content instance."""
+
+    def __init__(self, matcher, annotate=True):
+        super(AsText, self).__init__(
+            lambda log: log.as_text(), matcher, annotate=annotate)
+
+
+def throw(function, *args, **kwargs):
```
rbtcollins: Perhaps _raise or raise_ or something? My brain hurt when I read throw :/. Just a thought.

rbtcollins:
Two really minor nits; I think its good to go, with or without addressing them.

PR #167, opened 2015-11-19
Become compatible with newer Fixtures _setUp() API.

bigjools:
Previously, when gathering details caused by a setUp failure,
a traceback occurred if the fixture used the newer _setUp().
This also had the side effect of not clearing up fixtures properly.

Fixes: https://bugs.launchpad.net/testtools/+bug/1469759

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/167)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -19,6 +19,7 @@ The testtools authors are:
  * Vincent Ladeuil
  * Nikola Đipanov
  * Tristan Seligmann
+ * Julian Edwards
 
 and are collectively referred to as "testtools developers".
 
--- a/NEWS
+++ b/NEWS
@@ -21,6 +21,10 @@ Improvements
   report the lack of result as a test error. This ought to make weird
   concurrency interaction bugs easier to understand. (Jonathan Lange)
 
+* Previously, when gathering details caused by a setUp() failure,
+  a traceback occurred if the fixture used the newer _setUp().
+  This had the side effect of not clearing up fixtures nor gathering details
+  properly.  This is now fixed.  (Julian Edwards #1469759)
 
 2.0.0
 ~~~~~
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -685,7 +685,8 @@ class TestCase(unittest.TestCase):
         try:
             fixture.setUp()
         except MultipleExceptions as e:
-            if e.args[-1][0] is fixtures.fixture.SetupError:
+            if (fixtures is not None and
+                    e.args[-1][0] is fixtures.fixture.SetupError):
                 gather_details(e.args[-1][1].args[0], self.getDetails())
             raise
         except:
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -26,6 +26,7 @@ from extras import (
     try_import,
     try_imports,
     )
+fixtures = try_import('fixtures')
 # To let setup.py work, make this a conditional import.
 unittest = try_imports(['unittest2', 'unittest'])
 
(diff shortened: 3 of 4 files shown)


bigjools:
I initially removed the whole try/except block but it kills support for the older Fixtures setUp() API. I can change it back if you want to do that. However, inspecting private properties on the fixture isn't so great either.

bigjools:
Ping, is anyone looking at this please?

thomir:
Hi,

I suspect the reason this hasn't had the attention it deserves is due to the failing travis run. @rbtcollins Fixed that, so I guess we need to re-trigger a travis run. I'm not sure how to do that, perhaps Robert has an idea?

rbtcollins:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[testtools/testcase.py, line 675 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/167#-K45XMPnq3SqMXPksMyx:-K45XMPnq3SqMXPksMyy:1989936992) ([raw file](https://github.com/testing-cabal/testtools/blob/6dc87af1073e7ffbb25f30fa10f0d1941b4da5ea/testtools/testcase.py#L675)):*
This should be 'if safe_hasattr(fixture, '_details') and fixture._details is not None', since we can't trust any state in the failed fixture.

You also need to thread something like https://github.com/testing-cabal/fixtures/blob/master/fixtures/fixture.py#L258 in here, because we're not capturing the fixtures in the new exception either.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/167)*

bigjools:
Reviewable is confusing me. Commenting here instead.

"You also need to thread something like https://github.com/testing-cabal/fixtures/blob/master/fixtures/fixture.py#L258 in here, because we're not capturing the fixtures in the new exception either."

Sorry, I don't understand what you mean.

bigjools:
Ok I think I addressed all the concerns, let me know how it looks.

rbtcollins:
Very close to mergable. One little tweak needed, plus please add an entry to NEWS and yourself to AUTHORS if you are not already there.

---

Reviewed 2 of 2 files at r4.

---

*[testtools/testcase.py, line 685 \[r4\]](https://reviewable.io:443/reviews/testing-cabal/testtools/167#-K9B3fqoPqdrqPRIFBxB:-K9B3fqoPqdrqPRIFBxC:-554928261) ([raw file](https://github.com/testing-cabal/testtools/blob/15188a128fd5ad44b33be126226c7440e7e6b4cf/testtools/testcase.py#L685)):*
You need to either unconditionally import fixtures, or guard this with 'fixtures is not None and'

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/167#-:-K9B3xPHLjE3PoHTbuXn:-1347546708)*

bigjools:
I thought about guarding the conditional fixtures import before, but it's impossible to even be in that code path without it. I've added the guard anyway as there's nothing to lose.

Thanks for reviewing, let me know how it looks now!

rbtcollins:
Reviewed 3 of 3 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/167)*

rbtcollins:
Merged

PR #170, opened 2015-12-02
Latest Twisted doesn't run on Python 2.6

jml:
Travis tests are failing because the latest Twisted doesn't support Python 2.6. This removes such tests from our .travis.yml.

I don't _think_ this change is NEWS-worthy, but could be convinced otherwise.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/170)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,6 @@ matrix:
   include:
     - python: "2.6"
       env: TWISTED_REQ="Twisted==13.0.0"
-    - python: "2.6"
-      env: TWISTED_REQ="Twisted"
     - python: "2.7"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,8 +19,6 @@ matrix:
   include:
     - python: "2.6"
       env: TWISTED_REQ="Twisted==13.0.0"
-    - python: "2.6"
-      env: TWISTED_REQ="Twisted"
```
rbtcollins: Why not set this to the highest release before they dropped 2.6?

jml: Done.

jml:
Previous build failure was spurious. Merging now.

PR #171, opened 2015-12-06
Decouple Twisted logging inside AsynchronousDeferredRunTest

jml:
This patch works towards fixes for https://pad.lv/942785 and https://pad.lv/1515362. I believe it makes no behavioral changes and introduces no new APIs.

On IRC, @rbtcollins suggest that `AsynchronousDeferredRunTest` was conflating responsibilities: taking over the Twisted logs in addition to being responsible for spinning the reactor and gathering errors. 

This patch begins to untangle these responsibilities. It deprecates `run_with_log_observers` and ceases to use it. The functionality of `run_with_log_observers` is split into two Fixtures: `_NoTwistedLogObservers` (removes all log observers) and `_TwistedLogObservers` (adds some log observers).

It then provides two new specialized fixtures: `_CaptureTwistedLogs` (stores all Twisted logs as a detail) and `_ErrorObserver` (records all `log.err` calls—necessary for preserving expected behavior in Twisted tests). These fixtures are then used in `AsynchronousDeferredRunTest` to provide the same functionality that we provided before. 

My intent is that a follow-up patch would add parameters to `AsynchronousDeferredRunTest` to disable the use of `_NoTwistedLogObservers` and `_CaptureTwistedLogs`, and perhaps also expose the two fixtures at the same time. This would allow users to work around lp:942785 and lp:1515362. 

My big qualm about this patch is that it makes [fixtures](https://github.com/testing-cabal/fixtures) a dependency. Since testtools is a dependency of fixtures, this gives us a circular dependency. I think the "correct" fix is actually to move the entire `deferredruntest` option to a separate Python package (perhaps `txtesttools`), but even then, I don't know how we'd manage the deprecation smoothly. I welcome suggestions.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/171)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,7 +10,9 @@ Changes
 -------
 
 * Add a new test dependency of testscenarios. (Robert Collins)
-  
+* Make ``fixtures`` a real dependency, not just a test dependency.
+  (Jonathan Lange)
+
 1.8.1
 ~~~~~
 
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,8 @@ Changes
 * Add a new test dependency of testscenarios. (Robert Collins)
 * Make ``fixtures`` a real dependency, not just a test dependency.
   (Jonathan Lange)
+* ``testtools.deferredruntest.run_with_log_observers`` is deprecated.
+  (Jonathan Lange)
 
 1.8.1
 ~~~~~
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pyrsistent
 python-mimeparse
 unittest2>=1.0.0
 traceback2
+fixtures
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ pyrsistent
 python-mimeparse
 unittest2>=1.0.0
 traceback2
-fixtures
+fixtures>=1.3.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,6 @@ classifier =
 
 [extras]
 test =
-  fixtures
   testscenarios
   unittest2>=1.1.0
 
--- a/testtools/deferredruntest.py
+++ b/testtools/deferredruntest.py
@@ -154,7 +154,8 @@ class _CaptureTwistedLogs(Fixture):
         logs = StringIO()
         full_observer = log.FileLogObserver(logs)
         self.useFixture(_TwistedLogObservers([full_observer.emit]))
-        self.addDetail(self.LOG_DETAIL_NAME, Content(UTF8_TEXT, logs.getvalue))
+        self.addDetail(self.LOG_DETAIL_NAME,
+                       Content(UTF8_TEXT, lambda: [logs.getvalue()]))
 
 
 def run_with_log_observers(observers, function, *args, **kwargs):
@@ -174,7 +175,8 @@ def run_with_log_observers(observers, function, *args, **kwargs):
 _log_observer = _LogObserver()
 
 
-# XXX: Should really be in python-fixtures
+# XXX: Should really be in python-fixtures.
+# See https://github.com/testing-cabal/fixtures/pull/22.
 class _CompoundFixture(Fixture):
     """A fixture that combines many fixtures."""
 
@@ -384,6 +386,8 @@ class AsynchronousDeferredRunTest(_DeferredRunTest):
 
         # We can't just install these as fixtures on self.case, because we
         # need the clean up to run even if the test times out.
+        #
+        # See https://bugs.launchpad.net/testtools/+bug/897196.
         with logging_fixture as capture_logs:
             for name, detail in capture_logs.getDetails().items():
                 self.case.addDetail(name, detail)
--- a/testtools/deferredruntest.py
+++ b/testtools/deferredruntest.py
@@ -174,6 +174,19 @@ def run_with_log_observers(observers, function, *args, **kwargs):
 _log_observer = _LogObserver()
 
 
+# XXX: Should really be in python-fixtures
+class _CompoundFixture(Fixture):
+    """A fixture that combines many fixtures."""
+
+    def __init__(self, fixtures):
+        super(_CompoundFixture, self).__init__()
+        self._fixtures = fixtures
+
+    def _setUp(self):
+        for fixture in self._fixtures:
+            self.useFixture(fixture)
+
+
 def flush_logged_errors(*error_types):
     # XXX: jml: I would like to deprecate this in favour of
     # _ErrorObserver.flush_logged_errors so that I can avoid mutable global
@@ -342,27 +355,21 @@ class AsynchronousDeferredRunTest(_DeferredRunTest):
         # XXX: We want to make the use of these _NoTwistedLogObservers and
         # _CaptureTwistedLogs optional, and ideally, make it so they aren't
         # used by default.
+        logging_fixture = _CompoundFixture(
+            [_NoTwistedLogObservers(), _CaptureTwistedLogs()])
 
         # We can't just install these as fixtures on self.case, because we
         # need the clean up to run even if the test times out.
-        with _NoTwistedLogObservers():
-            # Can't use with for _CaptureTwistedLogs, because details are gone
-            # after cleanup.
-            capture_logs = _CaptureTwistedLogs()
-            capture_logs.setUp()
-            try:
-                with _ErrorObserver(_log_observer) as error_fixture:
-                    successful, unhandled = self._blocking_run_deferred(
-                        spinner)
-                for logged_error in error_fixture.flush_logged_errors():
-                    successful = False
-                    self._got_user_failure(
-                        logged_error, tb_label='logged-error')
-            finally:
-                for name, detail in capture_logs.getDetails().items():
-                    self.case.addDetail(name, detail)
-                capture_logs.cleanUp()
-
+        with logging_fixture as capture_logs:
+            for name, detail in capture_logs.getDetails().items():
+                self.case.addDetail(name, detail)
+            with _ErrorObserver(_log_observer) as error_fixture:
+                successful, unhandled = self._blocking_run_deferred(
+                    spinner)
+            for logged_error in error_fixture.flush_logged_errors():
+                successful = False
+                self._got_user_failure(
+                    logged_error, tb_label='logged-error')
 
         if unhandled:
             successful = False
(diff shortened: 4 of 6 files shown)


rbtcollins:
I haven't reviewed yet, shall do today. However I don't believe the circular dependency is a problem: pure Python has no build-time requirements, all our source can be compiled. We can't *test* until all things are compiled, and for distros that conflate build-and-test this is a problem, but one which is describable in the Debian multi-stage mechanism they added to deal with genuine loops during bootstrapping archives.

rbtcollins:
Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[testtools/deferredruntest.py, line 103 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tReBX8S8juu9eq9An:-K4tReBX8S8juu9eq9Ao:1977577462) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L103)):*
It would be safer to add this cleanup before calling _remove_observers. If _remove_observers is interrupted halfway through, no cleanups would run with the current code. Since _remove_observers itself isn't atomic, I'd actually not use it at all. Instead:

```
for observer in real_observers:
    publisher.removeObserver(observer)
    self.addCleanup(publisher.addObserver, observer)
```

Which should only lose observer if a signal based interrupt like KeyboardInterrupt is generated between those two lines of code.

---

*[testtools/deferredruntest.py, line 117 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tTbp521ajxvqTWW1n:-K4tTbp521ajxvqTWW1o:206547425) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L117)):*
And here too, naturally. Probably worth a single helper that takes publisher, observers, fixture

---

*[testtools/deferredruntest.py, line 129 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tTjXJeFCkp-brFeMx:-K4tTjXJeFCkp-brFeMy:1864731289) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L129)):*
Since this is only relevant when the fixture is setUp, I would either not set it here, or set it to None.

---

*[testtools/deferredruntest.py, line 137 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tU2uRBdCSZ3hk47Ps:-K4tU2uSBTAC5uCAUqIf:1750496285) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L137)):*
This is very odd: its making multiple use of the fixture hard/buggy to use - since its propogating state across uses. What is it for?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/171)*

jml:
Thanks for the review.

After reflecting over-night, I think even more that moving the Twisted code out of testtools into a separate package is the right thing to do. Even separately from the circular dependency issue, I think it's a clearer separation of responsibilities.

But maybe it's OK to land this as-is and see what happens.

---

Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---

*[testtools/deferredruntest.py, line 103 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tReBX8S8juu9eq9An:-K4vrpphUowS3ZymWi9I:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L103)):*
Done.

---

*[testtools/deferredruntest.py, line 117 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tTbp521ajxvqTWW1n:-K4vrunSx5-cKY6Tt379:-1388591041) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L117)):*
Fixed. Not made the helper, since each one is reversed from the other, and the helper would be more complicated than code it helps.

---

*[testtools/deferredruntest.py, line 129 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tTjXJeFCkp-brFeMx:-K4vsGVfkCmNWatNGgdr:1340379618) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L129)):*
Entire attribute gone now in response to comment below.

---

*[testtools/deferredruntest.py, line 137 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K4tU2uRBdCSZ3hk47Ps:-K4vuYulWWNxhrO9dOdS:419347317) ([raw file](https://github.com/testing-cabal/testtools/blob/25dc08a4b218d58409e3887129a0a0052f608920/testtools/deferredruntest.py#L137)):*
It's to capture errors logged while the fixture is active. 

I've ditched `logged_errors` and this method entirely, in favour of `_run_core` calling `flush_logged_errors` directly, which I like much better. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-:-K4vqeLhPFhG7G1YtMYk:-38669054)*

jml:
We discovered https://clusterhq.atlassian.net/browse/FLOC-3681 while using this patch. 

If we install `_NoTwistedLogObservers` as a fixture, and the test times out, we don't clean it up (see [#897196](https://bugs.launchpad.net/testtools/+bug/897196)). This means that tests that time out will fail with a weird error.

I've adjusted the code to use `with` for `_NoTwistedLogObservers` and a custom try/finally for `_CaptureTwistedLogs`.

---

Review status: 0 of 5 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-:-K5BhkAA7TRKkPp7E3x5:2092115452)*

rbtcollins:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[requirements.txt, line 9 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BcDTgLy5gKqckgJjY:-K9BcDTgLy5gKqckgJjZ:1349700068) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/requirements.txt#L9)):*
This will need a minimum version on it. Looks like 1.3.0 ?

---

*[testtools/deferredruntest.py, line 157 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9Bg6if9_2Kh0u6Cieh:-K9Bg6if9_2Kh0u6Ciei:577264319) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L157)):*
logs.getvalue returns a string, not a list of strings; this is going to be a potential performance problem with big logs - I recommend ```lambda: [logs.getvalue()]```

---

*[testtools/deferredruntest.py, line 187 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BgYR1L_qXJfodohuS:-K9BgYR1L_qXJfodohuT:-582829071) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L187)):*
I see you put up a PR for this; perhaps link to the PR? Or should we prioritise that ahead of this?

---

*[testtools/deferredruntest.py, line 386 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BgzMZrzjRpS7O8p66:-K9BgzMZrzjRpS7O8p67:-514707312) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L386)):*
That suggests a bug in the cleanup handling logic - and I saw the bug you filed about this. Perhaps reference the bug # here for your future self?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/171)*

jml:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions.

---

*[requirements.txt, line 9 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BcDTgLy5gKqckgJjY:-K9Blk5v7RRzITul_QXC:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/requirements.txt#L9)):*
Done.

---

*[testtools/deferredruntest.py, line 157 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9Bg6if9_2Kh0u6Cieh:-K9Blrlzkr6Vm2on4Eol:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L157)):*
Done.

---

*[testtools/deferredruntest.py, line 187 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BgYR1L_qXJfodohuS:-K9Bm9ZG4raygBt0nnJS:2092198634) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L187)):*
Done.

No need to prioritize that ahead of this. If https://github.com/testing-cabal/fixtures/pull/22 gets in, we can just delete this (it's private, no compat issues) and use it. Making something fit for the fixtures library is a different, higher standard than making something we can use here & now.

---

*[testtools/deferredruntest.py, line 386 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/171#-K9BgzMZrzjRpS7O8p66:-K9Bmo6LrTfKhTGBDXsk:1863749219) ([raw file](https://github.com/testing-cabal/testtools/blob/2428a78835ff2cb90f228a54b6691de21dd8eba2/testtools/deferredruntest.py#L386)):*
Added reference to https://bugs.launchpad.net/testtools/+bug/897196.

FWIW, I think the longer-term correct strategy is for `RunTest` to have its own separate, pluggable fixture.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/171)*

jml:
Squashed to master:  af8e294

PR #172, opened 2015-12-07
Make suppressing and storing Twisted logs optional

jml:
Fixes https://bugs.launchpad.net/testtools/+bug/1515362 and https://bugs.launchpad.net/testtools/+bug/942785. Depends on #171.

Adds two parameters to `AsynchronousDeferredRunTest` that allow callers to decide whether we suppress Twisted logs and whether we capture Twisted logs. Also exposes `CaptureTwistedLogs` so that users who want to disable the automatic capturing can use the fixture to do it themselves, thus allowing post-processing of logs information.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/172)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -9,6 +9,14 @@ Next
 Changes
 -------
 
+* ``AsynchronousDeferredRunTest`` now has ``suppress_twisted_logging`` and
+  ``store_twisted_logs`` parameters that can be used to override the default
+  logging behaviour.  (Jonathan Lange, #942785)
+
+* New fixture ``CaptureTwistedLogs`` that can be used with
+  ``AsynchronousDeferredRunTest`` to attach a detail containing everything
+  logged to Twisted during the test run.  (Jonathan Lange, #1515362)
+
 * Python 2.6 and 3.2 are no longer supported. If you want to use either of
   these versions of Python, use testtools 1.9.0. (Jonathan Lange)
 
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -30,3 +30,4 @@ testtools.twistedsupport
 
 .. automodule:: testtools.twistedsupport
    :members:
+   :special-members: __init__
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -984,6 +984,7 @@ exception or mangle the display.  In most cases simply avoiding the ``%s``
 format specifier and using ``%r`` instead will be enough.  For examples of
 more complex formatting see the ``testtools.matchers`` implementatons.
 
+.. _details:
 
 Details
 =======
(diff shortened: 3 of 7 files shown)


rbtcollins:
I wonder if you might rebase this slightly - you have a lot of exploratory and feature-branch-merge noise in the history, and it makes reviewing whats happened in e.g. a release quite a bit harder :(.  If it was bzr or loom, there'd be tooling to help :)

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/172#-:-K9Bjih7go-WfiwCAvLJ:708881559)*

jml:
Rebased. I can also split into two commits (one for exposing `CaptureTwistedLogs`, one for the rest) if you'd like.

Took me a little while fiddling w/ reviewable.io to make it show the right diffs.

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/172#-:-K9C8D8JmDlc91deosje:1822566517)*

NEWS
```diff
@@ -9,6 +9,14 @@ Next
 Changes
 -------
 
+* ``AsynchronousDeferredRunTest`` now has ``suppress_twisted_logging`` and
+  ``store_twisted_logs`` parameters that can be used to override the default
+  logging behaviour.  (Jonathan Lange, #942785)
+
+* New fixture ``CaptureTwistedLogs`` that can be used with
+  ``AsynchronousDeferredRunTest`` to attach a detail containing everything
```
thomir: If I'm not mistaken, CaptureTwistedLogs can be used with any test runner? If so, suggest removing 'with ``Async...``' as it sort of implies that it's only useful in that case. But perhaps I've misunderstood....

rbtcollins: Its a twisted specifical log capturerer, but not specific to testtools; thats true. OTOH testtools is currently the only twisted and details supporting runner; I think being overly general will make the changelog harder to understand.

Which reminds me: docs please; shouldn't this be in the for-test-authors docs?

jml:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[NEWS, line 17 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/172#-K4wZdHtocTumUdcLNTv-r2-17:-K9YqAf2G-0d3Oze0ZoB:34595947) ([raw file](https://github.com/testing-cabal/testtools/blob/1e2aa0d89f84718e6bcc959e68c33c54d69afbdf/NEWS#L17)):*
It should be in `twisted-support.rst`. OK for me to land after adding those, or should I get another round of review?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/172)*

rbtcollins:
Land away imo

jml:
Merged in  302751e

PR #175, opened 2015-12-09
Add getUniqueText creation method to TestCase

brantlk:
This adds a new creation method to TestCase for py2/3 to help application developers test for proper unicode handling.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/175)


rbtcollins:
We don't want to add things like this to TestCase, but we can certainly accept a patch to add the feature as a Fixture, or a generator or some such independent object.

brantlk:
ok, will look into a generator. There's no setup/cleanup required so fixture doesn't seem right.

PR #176, opened 2015-12-09
Add generate_unique_text generator

brantlk:
A generator is provided that generates unique (unicode) text strings. This is handy for tests that would prefer to use unicode strings in order to validate that their code works properly with unicode characters.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/176)

### DIFF
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -1346,9 +1346,9 @@ fine.  However, sometimes it's useful to be able to create arbitrary objects
 at will, without having to make up silly sample data.
 
 To help with this, ``testtools.TestCase`` implements creation methods called
-``getUniqueString``, ``getUniqueInteger``, and ``getUniqueText``.  They return
-strings and integers that are unique within the context of the test that can be
-used to assemble more complex objects.  Here's a basic example where
+``getUniqueString`` and ``getUniqueInteger``.  They return strings and integers
+that are unique within the context of the test that can be used to assemble
+more complex objects.  Here's a basic example where
 ``getUniqueString`` is used instead of saying "foo" or "bar" or whatever::
 
   class SomeTest(TestCase):
@@ -1385,9 +1385,9 @@ details of certain variables don't actually matter.
 See pages 419-423 of `xUnit Test Patterns`_ by Gerard Meszaros for a detailed
 discussion of creation methods.
 
-``getUniqueText`` is similar to ``getUniqueString``, except it generates text
-that contains unicode characters. As such the value will be a unicode string
-in Python2 and a str in Python3.
+``testcase.generate_unique_text()`` is similar to ``getUniqueString``, except
+it generates text that contains unicode characters. As such the value will be a
+unicode string in Python2 and a str in Python3.
 
 Test attributes
 ---------------
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -1346,9 +1346,9 @@ fine.  However, sometimes it's useful to be able to create arbitrary objects
 at will, without having to make up silly sample data.
 
 To help with this, ``testtools.TestCase`` implements creation methods called
-``getUniqueString`` and ``getUniqueInteger``.  They return strings and integers
-that are unique within the context of the test that can be used to assemble
-more complex objects.  Here's a basic example where
+``getUniqueString`` and ``getUniqueInteger``.  They return strings and
+integers that are unique within the context of the test that can be used to
+assemble more complex objects.  Here's a basic example where
 ``getUniqueString`` is used instead of saying "foo" or "bar" or whatever::
 
   class SomeTest(TestCase):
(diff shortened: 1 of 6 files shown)


doc/for-test-authors.rst
```diff
@@ -1385,6 +1385,10 @@ details of certain variables don't actually matter.
 See pages 419-423 of `xUnit Test Patterns`_ by Gerard Meszaros for a detailed
 discussion of creation methods.
 
+``testcase.generate_unique_text()`` is similar to ``getUniqueString``, except
+it generates text that contains unicode characters. As such the value will be a
+unicode string in Python2 and a str in Python3.
```
rbtcollins: This is confusing as written, perhaps:

The value will be a ``unicode`` object in Python 2, and a ``str`` object in Python 3.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pyrsistent
 python-mimeparse
 unittest2>=1.0.0
 traceback2
+six
```
rbtcollins: What version of six introduced the functions you're using? I've seen bug reports galore with mock due to folk having old distro copies installed :(

testtools/testcase.py
```diff
@@ -173,6 +175,49 @@ def gather_details(source_dict, target_dict):
...
+    return s
+
+
+def unique_text_generator(prefix):
+    """Generates unique text values.
+
+    Generates text values that are unique. Use this when you need arbitrary
+    text in your test, or as a helper for custom anonymous factory methods.
+
+    :param prefix: The prefix for text. If this is a TestCase the id is used.
```
rbtcollins: Uhm, eeek. I'd rather not type-sniff: just have callers pass in self.id() if they want to use their id as a prefix.

testtools/testcase.py
```diff
@@ -173,6 +175,49 @@ def gather_details(source_dict, target_dict):
...
+        prefix = prefix.id()
+
+    index = 0
+
+    # 0x1e00 is the start of a range of chars that are easy to see are
+    # unicode since they've got circles and dots and other diacriticals.
+    # 0x1eff is the end of the range of these diacritical chars.
+    BASE_CP = 0x1e00
+    CP_RANGE = 0x1f00 - BASE_CP
+
```
rbtcollins: I'd prefer a little less vertical whitespace in here - (per PEP8), since the function really is one logical thing.

testtools/testcase.py
```diff
@@ -173,6 +175,49 @@ def gather_details(source_dict, target_dict):
...
+    while True:
+        unique_text = _unique_text(BASE_CP, CP_RANGE, index)
+        yield six.text_type('%s-%s') % (prefix, unique_text)
+        index = index + 1
```
rbtcollins: since index is getting pushed through mod, is this actually unique?

rbtcollins:
Please also add a NEWS entry, and import the new thing into testtools/__init__.py, and export it in testtools/__init__.py's __all__.

doc/for-test-authors.rst
brantlk: Done.

requirements.txt
brantlk: six 1.4.0 and later have the functions.

testtools/testcase.py
brantlk: Removed the extra code.

testtools/testcase.py
brantlk: Done.

testtools/testcase.py
brantlk: You'll always get a new value each time .next() is called on the result of unique_text_generator(). The sequence is like A, B, ... Z, AA, AB, ..., AZ, etc. (where A is 0x1e00 and Z is 0x1eff).

testtools/testcase.py
```diff
@@ -21,6 +22,8 @@
 import sys
 import types
 
+import six
+
```
rbtcollins: six is a third party import, so it should be grouped with the extras and unittest2 imports - but i see we have a little odd ordering here already, - perhaps put it with the wraps import below?

brantlk: Done

rbtcollins:
ping (this is failing tests, i haven't looked into why)

brantlk:
Python 3.2 doesn't support string literals like u'\u1e00'. https://docs.python.org/3.2/reference/lexical_analysis.html#literals

rbtcollins:
We've dropped 3.2 support for the next release, so thats ok - but the way we address that in other code was the u() function.

u("\u1e00")

brantlk:
I see there's a _u() function. Hopefully it'll work (I wasn't able to virtualenv with python 3.2 on my system so can't try it out).

rbtcollins:
Rebased, fixed one little PEP8 quirk and merged.

PR #177, opened 2015-12-10
Enable as_text() for JSON

jml:
It's super useful to be able to attach a chunk of JSON data to a test and then see that data when the test fails. 

This approach has its defects (e.g. it doesn't pretty print the JSON), but maybe it's OK, and a patch is a great way to start a conversation.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/177)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,6 +12,7 @@ classifier =
 [extras]
 test =
   fixtures
+  hypothesis
   testscenarios
   unittest2>=1.1.0
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,6 @@ classifier =
 [extras]
 test =
   fixtures
-  hypothesis
   testscenarios
   unittest2>=1.1.0
 
--- a/testtools/content.py
+++ b/testtools/content.py
@@ -104,7 +104,7 @@ class Content(object):
 
         :raises ValueError: If the content type is not text/\*.
         """
-        if self.content_type.type != "text":
+        if not self.content_type.is_text():
             raise ValueError("Not a text type %r" % self.content_type)
         return self._iter_text()
 
--- a/testtools/content_type.py
+++ b/testtools/content_type.py
@@ -42,6 +42,7 @@ class ContentType(object):
         return self.type == 'text' or self._text_override
 
 
-JSON = ContentType('application', 'json', text_override=True)
+JSON = ContentType(
+    'application', 'json', {'charset': 'utf8'}, text_override=True)
 
 UTF8_TEXT = ContentType('text', 'plain', {'charset': 'utf8'})
--- a/testtools/content_type.py
+++ b/testtools/content_type.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2009-2012 testtools developers. See LICENSE for details.
+# Copyright (c) 2009-2015 testtools developers. See LICENSE for details.
 
 """ContentType - a MIME Content Type."""
 
@@ -34,6 +34,10 @@ class ContentType(object):
         else:
             params = ''
         return "%s/%s%s" % (self.type, self.subtype, params)
+
+    def is_text(self):
+        """True if this content type can be converted to text."""
+        return self.type == 'text'
 
 
 JSON = ContentType('application', 'json')
--- a/testtools/content_type.py
+++ b/testtools/content_type.py
@@ -12,7 +12,8 @@ class ContentType(object):
         content type.
     """
 
-    def __init__(self, primary_type, sub_type, parameters=None):
+    def __init__(self, primary_type, sub_type, parameters=None,
+                 text_override=False):
         """Create a ContentType."""
         if None in (primary_type, sub_type):
             raise ValueError("None not permitted in %r, %r" % (
@@ -20,6 +21,7 @@ class ContentType(object):
         self.type = primary_type
         self.subtype = sub_type
         self.parameters = parameters or {}
+        self._text_override = text_override
 
     def __eq__(self, other):
         if type(other) != ContentType:
@@ -37,9 +39,9 @@ class ContentType(object):
 
     def is_text(self):
         """True if this content type can be converted to text."""
-        return self.type == 'text'
+        return self.type == 'text' or self._text_override
 
 
-JSON = ContentType('application', 'json')
+JSON = ContentType('application', 'json', text_override=True)
 
 UTF8_TEXT = ContentType('text', 'plain', {'charset': 'utf8'})
--- a/testtools/tests/test_content.py
+++ b/testtools/tests/test_content.py
@@ -212,6 +212,13 @@ class TestContent(TestCase):
         expected = Content(JSON, lambda: [_b('{"foo": "bar"}')])
         self.assertEqual(expected, json_content(data))
 
+    def test_json_content_as_text(self):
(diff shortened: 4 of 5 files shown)


jml:
Another thing I thought of was that maybe `ContentType` could be responsible for figuring out how to convert to text. It sort of makes sense, but I like it being a dumb object.

bz2:
Basic idea here seems fine, lots of content types don't start text/ but are an ascii/utf-8 subset so are safe to convert to text. The main issue is some have a different mechanism of determining what the charset actually is. So, you can't just treat application/xml as text without implementing the logic over charset detection.

The more comprehesive option seems like adding something like a mapping of content types to conversion functions in content.py that could say decode `JSON` with `json.loads()` then encode as text with `json.dumps(o, indent=4, sort_keys=True).decode('utf-8')` or something.

---

Reviewed 4 of 4 files at r1.

---

*[testtools/content_type.py, line 45 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bwdzTwf-Ts7RLNguj:-K5bwdzUfBuiwzu2c5jX:-142443500) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/content_type.py#L45)):*
You'd also have to put `{'charset': 'utf8'}` in here, otherwise it gets default decoded as latin-1 by `Content._iter_text()`?

---

*[testtools/tests/test_content.py, line 218 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bx4owloD5Yl3QOK8H:-K5bx4owloD5Yl3QOK8I:26657765) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/tests/test_content.py#L218)):*
Throw in some non-ascii data.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-:-K5bwEInBCg4sWGLr-Xf:767355205)*

jml:
Thanks for the review, much appreciated!

Just to be clear, are you asking that I implement this more comprehensive solution as part of this PR?

---

Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[testtools/content_type.py, line 45 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bwdzTwf-Ts7RLNguj:-K5ePaAiON1qJ4QomMZ_:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/content_type.py#L45)):*
Done.

---

*[testtools/tests/test_content.py, line 218 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bx4owloD5Yl3QOK8H:-K5eOWuw26pCx7RrMT5W:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/tests/test_content.py#L218)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-:-K5eONYudqeNH11jcIo1:-688928497)*

bz2:
Just this PR is fine. Not all mime types will require a fancy path anyway.

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[testtools/content_type.py, line 45 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bwdzTwf-Ts7RLNguj:-K5jo2YTpS8VUZhC0tkG:322553975) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/content_type.py#L45)):*
Hmm, we don't use the ContentType repr anywhere important right? Technically application/json doesn't have a charset, but this is a handy way of making the code work. The alternative would be making text_override set a back-channel charset value rather than True. Your call.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-:-K5jmewahKUzX3wjRbaz:2139656095)*

rbtcollins:

---

*[testtools/content_type.py, line 45 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bwdzTwf-Ts7RLNguj:-K5l4J3c5n4GuyPC8cKW:-1071838324) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/content_type.py#L45)):*
The charset thing is a problem... https://tools.ietf.org/html/rfc7159 - application/json may be utf8/16/32 - section 8.1 "The default is utf8". And sadly the RFC still does not provide a means to determine encoding, other than trying them all :/.

I wonder if a registry of decoders would be a better strategy than teaching ContentType's code about which non-text types are usable as text. 

I certainly get the use case here: one has some JSON, and would like to show it, but still highlight it as being JSON to other consumers than the CLI.

Relatedly http://tools.ietf.org/html/rfc6657 updates things in this space a bit too.

---

*[testtools/tests/test_content.py, line 220 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5l90GBUvQ-NxBCmlAa:-K5l90GBUvQ-NxBCmlAb:1598451369) ([raw file](https://github.com/testing-cabal/testtools/blob/85f9cde25777b07028c146f681204b81ae61adc4/testtools/tests/test_content.py#L220)):*
So, we do actually depend on the repr of ContentType in subunit - I'd like to know that this patch does not make it non-conforming to the RFC. I think it does.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/177)*

jml:
Thanks. Since @rbtcollins has also said that a more comprehensive approach is actually needed, I'm going to cancel this PR. It's different enough that very little from this would actually be useful.

---

---

*[testtools/content_type.py, line 45 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5bwdzTwf-Ts7RLNguj:-K5ujn5_OHsUJHrRmzJ1:-920445207) ([raw file](https://github.com/testing-cabal/testtools/blob/506f809dd6876afade6d453ec6ce53422418b6b7/testtools/content_type.py#L45)):*
OK. I'll explore a registry of decoders option in another PR. I _really_ want to avoid having more global mutable state, so I think I'll do it as an explicit parameter.

That approach also has the advantage that it doesn't actually have to be in testtools.

---

*[testtools/tests/test_content.py, line 220 \[r2\]](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-K5l90GBUvQ-NxBCmlAa:-K5ujd_TlgbaE_B0Dp_W:1896665674) ([raw file](https://github.com/testing-cabal/testtools/blob/85f9cde25777b07028c146f681204b81ae61adc4/testtools/tests/test_content.py#L220)):*
Is this comment related to the code above it?

I suggest you add a test for this condition in a separate PR.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/177#-:-K5uk938qGKjZFOhtR1R:-1130358167)*

PR #179, opened 2015-12-14
Add map_bytes for Content

jml:
Makes `Content` a functor, which can be handy for some operations.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/179)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,8 +11,9 @@ Changes
 
 * Add a new test dependency of testscenarios. (Robert Collins)
 
-* New method ``Content.map_bytes`` that can be used to return a new ``Content``
-  object of the same type with data based on the original. (Jonathan Lange)
+* New function ``map_bytes`` that can be used to map a ``Content`` object to a
+  new ``Content`` of the same content type with data based on the original.
+  (Jonathan Lange)
 
 1.8.1
 ~~~~~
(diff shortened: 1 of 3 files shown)


jml:
I want a sanity check before adding a new API, so not merging yet.

rbtcollins:
This is an ok thing to do.

Alternative option. If we made Content have an __iter__ (that yields bytes), this could be spelt by users as:
c = Content(base.content_type, lambda: map(base, transform))

hmmm, well it can be spelt that way by users already I guess.

I wonder if we're going to dig ourselves a naming consistency hole here though - since this is the first of this sort of thing we've got. 

That said, this isn't currently exported (in __all__ in the root and down to the place its defined), so I'd rather it either be exported, or be _ prefixed, to make its status crystal clear.

jml:
Exported from testtools.content (we don't presently export anything from testtools.content further out—I believe our historical thinking was that it might want to become its own package).

Separately from this PR, we should document the conventions around exactly what is & isn't public and thus supported.

jml:
@rbtcollins What's the resolution on this PR? Go or no-go?

jml:
Don't want this on my Github dashboard. Others are free to re-submit if they'd like.

rbtcollins:
Ugh, sorry.

PR #180, opened 2015-12-15
Support Python 3.5, announce deprecation of Python 3.2.

jml:
This patch adds Python 3.5 to the documented supported Pythons, and to the Travis file. It also announces that this will be the last release to support Python 3.2.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/180)


jml:
This is a moderately big deal, so waiting on review, even though timeout has elapsed.

PR #181, opened 2015-12-19
Interfaces for Matchers

jml:
We talked about having explicit interfaces for things. This PR has come out of my various experiments for adding such things. It adds `IMatcher` and `IMismatch` interfaces, makes the documentation and call sites refer to them and updates the interface tests to use them.

It also updates the filesystem matcher tests to actually _use_ the interface tests.

I made the call that `IMismatch.describe` should return text, not bytes, as it seems like the right type, and  doesn't impose an onerous burden for implementers. All of the code that handles bytes is still there, but I've tried put deprecation warnings in wherever someone might pass bytes, or might pass their own matcher that returns a byte-returning mismatcher.

Much of the patch is updating all of the tests and all of the matchers to actually return unicode.

Future work:

* Add interfaces for `Content` so the matcher interface tests can be more thorough
* Add interface for `IDetailed` for objects that support adding and getting details 
* Add optional `detailed` parameter to `assert_that` standalone method so that it implements the full functionality of `TestCase.assertThat`
* Remove `Matcher` and `Mismatch` subclassing from all builtin matchers and mismatches

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/181)

### DIFF
--- a/testtools/compat.py
+++ b/testtools/compat.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2008-2015 testtools developers. See LICENSE for details.
 
-"""Compatibility support for python 2 and 3."""
+"""Compatibility support for Python 2 and 3."""
 
 __metaclass__ = type
 __all__ = [
@@ -21,9 +21,7 @@ import codecs
 import io
 import locale
 import os
-import re
 import sys
-import traceback
 import unicodedata
 
 from extras import try_import, try_imports
@@ -68,6 +66,7 @@ if sys.version_info > (3, 0):
         return (type,)
     str_is_unicode = True
     text_or_bytes = (str, bytes)
+    text = str
 else:
     import __builtin__ as builtins
     def _u(s):
@@ -86,6 +85,7 @@ else:
         return (type, types.ClassType)
     str_is_unicode = sys.platform == "cli"
     text_or_bytes = (unicode, str)
+    text = unicode
 
 _u.__doc__ = __u_doc
 
--- a/testtools/matchers/_impl.py
+++ b/testtools/matchers/_impl.py
@@ -70,6 +70,8 @@ class Mismatch(object):
             to the empty dict.
         """
         if description:
+            if not isinstance(description, text):
+                _text_deprecation(description)
             self._description = description
         if details is None:
             details = {}
--- a/testtools/matchers/_impl.py
+++ b/testtools/matchers/_impl.py
@@ -22,9 +22,10 @@ from testtools.compat import (
     _isbytes,
     istext,
     str_is_unicode,
-    text_repr
+    text_repr,
+    text,
     )
-from ._imatcher import IMatcher, IMismatch
+from ._imatcher import IMatcher, IMismatch, _text_deprecation
 
 
 @implementer(IMatcher)
@@ -121,11 +122,13 @@ class MismatchError(AssertionError):
         super(MismatchError, self).__init__()
         self.matchee = matchee
         self.matcher = matcher
-        self.mismatch = mismatch
+        self.mismatch = IMismatch(mismatch, mismatch)
         self.verbose = verbose
 
     def __str__(self):
         difference = self.mismatch.describe()
+        if not isinstance(difference, text):
+            _text_deprecation(difference)
         if self.verbose:
             # GZ 2011-08-24: Smelly API? Better to take any object and special
             #                case text inside?
@@ -159,9 +162,11 @@ class MismatchDecorator(object):
     def __init__(self, original):
         """Construct a `MismatchDecorator`.
 
-        :param original: A `Mismatch` object to decorate.
+        :param IMismatch original: A mismatch object to decorate. Will attempt
+            to adapt ``original`` to ``IMismatch``. If this is not possible,
+            will use ``original`` directly.
         """
-        self.original = original
+        self.original = IMismatch(original, original)
 
     def __repr__(self):
         return '<testtools.matchers.MismatchDecorator(%r)>' % (self.original,)
--- a/testtools/matchers/_impl.py
+++ b/testtools/matchers/_impl.py
@@ -17,14 +17,17 @@ __all__ = [
     'MismatchError',
     ]
 
+from zope.interface import implementer
 from testtools.compat import (
     _isbytes,
     istext,
     str_is_unicode,
     text_repr
     )
+from ._imatcher import IMatcher, IMismatch
 
 
+@implementer(IMatcher)
 class Matcher(object):
     """A pattern matcher.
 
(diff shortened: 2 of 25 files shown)


testtools/compat.py
```diff
@@ -86,6 +85,7 @@ def classtypes():
         return (type, types.ClassType)
     str_is_unicode = sys.platform == "cli"
     text_or_bytes = (unicode, str)
+    text = unicode
```
rbtcollins: I think we have a dependency on six now, so lets just use six.text_type rather than adding to this module?

testtools/matchers/_impl.py
```diff
@@ -117,11 +124,13 @@ def __init__(self, matchee, matcher, mismatch, verbose=False):
         super(MismatchError, self).__init__()
         self.matchee = matchee
         self.matcher = matcher
-        self.mismatch = mismatch
+        self.mismatch = IMismatch(mismatch, mismatch)
```
rbtcollins: So, I very much dislike this, as its referencing the global registry. That lets anyone break this code accidentally :(.

testtools/matchers/_impl.py
```diff
@@ -154,9 +164,11 @@ class MismatchDecorator(object):
     def __init__(self, original):
         """Construct a `MismatchDecorator`.
 
-        :param original: A `Mismatch` object to decorate.
+        :param IMismatch original: A mismatch object to decorate. Will attempt
+            to adapt ``original`` to ``IMismatch``. If this is not possible,
+            will use ``original`` directly.
         """
-        self.original = original
+        self.original = IMismatch(original, original)
```
rbtcollins: And here.

rbtcollins:
So here are my thoughts on this - based on the IRC chat and the patch.

I don't like the deprecation of 'str' descriptions in Python 2.7 - I don't see it buying us or our users anything other than strictness. We get that for free in 3.x, and a very large % of this patch is just scattering _u() around to accommodate this change. AFAICT its not required by having interfaces here. There is a minor thing, in that you've coded up deprecation logic, but the interface docstrings don't acknowledge this - if we do take this patch, I think its important to avoid user confusion by having these be internally consistent.

The z.i. adapter registry is in many ways an anti pattern that makes code harder to reason about, not better - glyph's blog post misses this in its praise for being able to pass in absolutely anything to a function using the global registry :).

AFAICT with the current patch, custom matchers doing their own mismatch and descriptions can be broken by this, because the patch no longer duck types. I don't think doing a major release is justified by the benefits of having better typing data - while I do want that typing data :). Update: you said on IRC that it falls back to x silently so isn't a break.

when mypy came out I did some experimentation with it, and got a bunch of bugs filed; they've mostly been fixed now, and I see from Guido's recent -ideas post that he's now keen on supporting Python 2.7 fully via comments (which is one of the things I raised, though I suspect thats lost in the mists of time :)) - so I think in short order we'll have mypy able to run sanely on testtools, if it doesn't already.

right now we already define the types - e.g. matchers.Mismatch - so we can allow folk to verify their custom matchers with mypy very very easily if we make sure that our code is annotated, or supply stubs ourselves - without requiring a second identical definition, or even abcs for most of our types. (Where things are abstract, abc is more relevant, but still perhaps overkill (and the global state thing again I don't like).

testtools/compat.py
jml: We don't actually, at least, not a direct one. Want me to _add_ a dependency on six as part of this PR?

testtools/matchers/_impl.py
jml: I'm convinced. Will change (if this PR goes ahead at all).

jml:
Thanks for taking the time to review this!

*str deprecation*

AIUI, the options are "don't do it" or "update the interface documentation to deprecate, not disallow". I infer that you'd prefer the former. I'm OK with that. 

*global adapter registry*

Agreed, let's not use it.

*mypy*

AFAICT, it's orthogonal to providing abstract types _and_ to providing adaptation support, so not 100% sure how it's relevant. I'm very happy to use PEP-484 syntax in docstrings though.

*concrete types*

> right now we already define the types - e.g. matchers.Mismatch - so we can allow folk to verify their custom matchers with mypy very very easily if we make sure that our code is annotated, or supply stubs ourselves - without requiring a second identical definition, or even abcs for most of our types. (Where things are abstract, abc is more relevant, but still perhaps overkill (and the global state thing again I don't like).

I disagree that abc is overkill. For me, the whole point of this is to provide abstract types, rather than concrete ones. For example: I'd like to use an immutable implementation of `Mismatch` that doesn't inherit from our mutable implementation. Having structured documentation of the mismatch protocol is extremely useful for doing such a thing. 

It's super important in reducing the amount of inheritance that we have, and in reducing (or at least, documenting) coupling. (e.g. https://github.com/jml/testtools/blob/interfaces/testtools/_itesttools.py#L99 documents what `RunTest` needs of a `TestCase`)

testtools/compat.py
rbtcollins: That would be fine; unittest2 has a hard dep, and we hard dep on that.

rbtcollins:
Lets discriminate between 'abc' and 'abstract types'.
```
class RunTestClient:
    def foo(self):
        raise NotImplementedError(self.foo)
```

is a cromulent abstract type which things can implement or not, or subclass or not, with no global ramifications.

abc is a registry based metaclass system allowing things to claim subclass status without sharing any code, or to cause the class definition to be broken if abstract methods are not implemented. Its also useful, but doesn't seem to be related to the things you want to accomplish.

I am confused though - Mismatch is immutable today (to the extent that anything in Python is) - none of its code mutates state, nor that of anything accessed via it. And nothing consuming it assumes mutability either.

jml:
> abc is a registry based metaclass system allowing things to claim subclass status without sharing any code

Yep. I want a system where things can claim to be the same type (fsvo 'type') without sharing code. Another way to approximate this is fully abstract classes (like the one you specified), but unlike `_BinaryComparison` (which some people would call an abstract type).

We can do that, although need to be clear in the documentation that we mean "something that implements this protocol" rather than "something for which `isinstance(x, Foo)` holds". 

> I am confused though - Mismatch is immutable today (to the extent that anything in Python is) - none of its code mutates state, nor that of anything accessed via it. And nothing consuming it assumes mutability either.

It's not documented as immutable, and as such any subclass can rely on changing state, perhaps by manipulating `_description` and `_details` (elsewhere in testtools, an underscore prefix means "public to subclasses").

If we changed it to use something like `pyrsistent.PClass`, e.g.

```python
class Mismatch(PClass):
  _description = field(text)
  _details = field(pmap)

  # ... methods here
```

Then almost all of the matchers in testtools would work fine, but some wouldn't. It'd be an API break, because we allow subclassing.

So… I think we're converging on fully-abstract-but-otherwise-vanilla classes?

jml:
Am going to try this again, but with plain old classes as the abstract types, and no type restriction.

PR #190, opened 2015-12-29
Prefer standard library SkipTest to unittest2.SkipTest

jml:
* `TestSkipped` ought to be deprecated, but we should wait until after we stop supporting Python 2.6, since people on Python 2.6 ought to have a version of testtools that isn't spamming them with warnings that we can't fix
* `unittest.SkipTest` is preferred to `unittest2.SkipTest`, so that by default we're using what other libraries are using

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/190)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -43,6 +43,10 @@ Changes
 * Report on all duplicate test ids when sorting test suites that contain
   duplicate ids.  (Thomas Bechtold, Jonathan Lange)
 
+* On Pythons greater than 2.6, ``TestSkipped`` will default to
+  ``unittest.SkipTest``, even if ``unittest2`` is installed.
+  (Jonathan Lange, #1529143)
+
 1.8.1
 ~~~~~
 
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2008-2011 testtools developers. See LICENSE for details.
+# Copyright (c) 2008-2015 testtools developers. See LICENSE for details.
 
 """Test case related stuff."""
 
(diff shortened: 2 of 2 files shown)


rbtcollins:
Never use the stdlib one imo. See unit test 2 innards.

jml:
Ugh. This isn't going to work, see https://travis-ci.org/jml/testtools/jobs/99235842. 

```
======================================================================
FAIL: testtools.tests.test_testsuite.TestConcurrentStreamTestSuiteRun.test_setupclass_skip
----------------------------------------------------------------------
Traceback (most recent call last):
  File "testtools/tests/test_testsuite.py", line 232, in test_setupclass_skip
    self.assertEqual(['addSkip'], [item[0] for item in log])
  File "testtools/testcase.py", line 363, in assertEqual
    self.assertThat(observed, matcher, message)
  File "testtools/testcase.py", line 448, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: ['addSkip'] != ['addError']
```

Current guess is that the right fix is to change unittest2 to prefer stdlib `SkipTest`.

jml:
> Never use the stdlib one imo. See unit test 2 innards.

Got this while posting my last comment. Have looked at [unittest2 innards](https://hg.python.org/unittest2/file/tip/unittest2/case.py) but still think that `unittest2.SkipTest` should subclass `unittest.SkipTest`. Closing this PR, because the fix for the problem of people expecting `unittest.SkipTest` for skips but getting something else can't be fixed while we delegate to unittest2, and I think we really ought to continue to delegate to unittest2.

PR #191, opened 2016-01-09
Add requirements for readthedocs

jml:
Currently, readthedocs is not including the API documentation for Twisted, since we don't declare Twisted as a hard dependency.

This fixes the issue by adding a requirements file specifically devoted to readthedocs. Separately, I've told readthedocs to use this requirements file.

Going to merge this right away, since that seems to be the only way to test it. :(

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/191)


jml:
If I push my branch in the main repo, I can test much more easily. Closing this now. Will create a new PR once I've got it working properly.

PR #192, opened 2016-01-09
Add requirements for readthedocs

jml:
Currently, readthedocs is not including the API documentation for Twisted, since we don't declare Twisted as a hard dependency.

This fixes the issue by adding a requirements file specifically devoted to readthedocs. Separately, I've told readthedocs to use this requirements file.

Before: http://testtools.readthedocs.org/en/latest/api.html#testtools-deferredruntest
After: http://testtools.readthedocs.org/en/readthedocs-twisted/api.html#module-testtools.deferredruntest

Now that I've verified that it works, I've reverted the readthedocs config (so that our regular docs will still build). Once I land this PR, I'll restore it (i.e. Set the requirements file to `readthedocs-requirements.txt` on https://readthedocs.org/dashboard/testtools/advanced/).

(Originally #191).

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/192)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -43,6 +43,9 @@ Changes
 * Report on all duplicate test ids when sorting test suites that contain
   duplicate ids.  (Thomas Bechtold, Jonathan Lange)
 
+* Add ``readthedocs-requirements.txt`` so readthedocs.org can build the
+  Twisted API documentation. (Jonathan Lange)
+
 1.8.1
 ~~~~~
 
--- a/readthedocs-requirements.txt
+++ b/readthedocs-requirements.txt
@@ -1,0 +1,1 @@
+Twisted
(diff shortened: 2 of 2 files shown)


rbtcollins:
Why the new file? I'd rather see this as an extra called e.g. docs

jml:
Me too, but readthedocs only does `python setup.py install` and only allows you to specify the path to a requirements.txt. I could add an extra and change the txt file to install `.[docs]` but don't know who that would help.

rbtcollins:
Sadness. The -e .[test] thing in the requirements file will work badly, because rtd does a non-inplace install subsequent to that.

testtools[test]
Twisted

should work though.

rbtcollins:
and perhaps a comment in the file about its intent

jml:
Done. I've verified that the documentation builds correctly.

Issue #93, opened 2014-08-24
SuccessResultOf matcher for synchronous Deferred testing


mithrandi:
I wrote this, based on `SynchronousTestCase.successResultOf` from Twisted:

```
class SuccessResultOf(object):
    """
    Match if a Deferred has fired with a result.
    """
    def __init__(self, resultMatcher=None):
        self.resultMatcher = resultMatcher


    def __str__(self):
        return 'SuccessResultOf(%s)' % (self.resultMatcher or '')


    def match(self, matchee):
        result = []
        matchee.addBoth(result.append)
        if not result:
            return Mismatch(
                'Success result expected on %r, found no result instead' % (
                    matchee,))
        elif isinstance(result[0], Failure):
            return Mismatch(
                'Success result expected on %r,'
                ' found failure result instead:\n%s' % (
                    matchee, result[0].getTraceback()))
        elif self.resultMatcher:
            return self.resultMatcher.match(result[0])
        return None
```

Presumably it would make sense to add FailureResultOf and NoResult matchers as well; the implementation of these would be a trivial permutation of the SuccessResultOf logic.

rbtcollins:
ping?

mithrandi:
Sorry, life happened ;) I haven't completely forgotten about this ticket, just got bumped a bit down my todo list.

jml:
FWIW, I'm working on this: https://github.com/jml/testtools/tree/deferred-matchers

mithrandi:
Merged in a5ea4b5365b8a2426e298be6b5182cfa529905fc.

PR #193, opened 2016-01-09
Add matchers for Deferreds

jml:
Addresses #93 by adding new matchers: `successful`, `failed`, and `no_result`.

This is ready for review, even if it's not 100% done.

Note:
* Shuffled some code around to clarify my own thinking
* Behaviour differs from Twisted in that `no_result` doesn't suppress the failure
* Lots of tweaks to the docs, which I've tested locally
* I'll add a NEWS entry once they are exported, see question below.
* No direct tests for `on_deferred_result` or `failure_content` because they are private and they were created during a "refactor" step

Questions:
* I think `no_result` should be made a constant. Agree?
* What do we think of the names?
* Where should I export them? I'm thinking of creating a new module `testtools.twisted` and exporting them there (along with `DeferredRunTest` and kin)
* Should `failed` take exception types? If so, can we defer to another branch?

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/193)


doc/conf.py
```diff
@@ -192,3 +195,8 @@
 
 # If false, no module index is generated.
 #latex_use_modindex = True
+
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/2', None),
```
rbtcollins: /2 ?

testtools/_deferred.py
```diff
@@ -0,0 +1,110 @@
...
+
+from testtools.content import TracebackContent
+
+
+class DeferredNotFired(Exception):
+    """Raised when we extract a result from a Deferred that's not fired yet."""
+
+    def __init__(self, deferred):
+        super(DeferredNotFired, self).__init__(
+            "%r has not fired yet." % (deferred,))
```
rbtcollins: TBH I'd prefer this on one line for readability. Or perhaps:

```
msg = "%r..."
super(DeferredNotFiled, self).__init__(msg)
```

testtools/_deferred.py
```diff
@@ -0,0 +1,110 @@
...
+    """Extract the result from a fired deferred.
+
+    It can happen that you have an API that returns Deferreds for
+    compatibility with Twisted code, but is in fact synchronous, i.e. the
+    Deferreds it returns have always fired by the time it returns.  In this
+    case, you can use this function to convert the result back into the usual
+    form for a synchronous API, i.e. the result itself or a raised exception.
+
+    It would be very bad form to use this as some way of checking if a
+    Deferred has fired.
```
rbtcollins: bad form? Will it cause bugs? I presume the issue is that the function mutates the deferred. Perhaps say that?

testtools/_deferred.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+    failures = []
+    successes = []
+    deferred.addCallbacks(successes.append, failures.append)
+    if len(failures) == 1:
+        failures[0].raiseException()
+    elif len(successes) == 1:
+        return successes[0]
+    else:
+        raise DeferredNotFired(deferred)
```
rbtcollins: Can failures or successes be > 1 ?

testtools/_deferred.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class ImpossibleDeferredError(Exception):
+    """Raised if a Deferred somehow triggers both a success and a failure."""
+
+    def __init__(self, deferred, successes, failures):
+        super(ImpossibleDeferredError, self).__init__(
```
rbtcollins: same legibility concern.

BTW for me a big chunk of this is the really weird left ) indenting where its not left aligned with the parameters.

testtools/_deferredmatchers.py
```diff
@@ -0,0 +1,188 @@
+# Copyright (c) testtools developers. See LICENSE for details.
+
+"""Matchers that operate on synchronous Deferreds.
+
+A "synchronous" Deferred is one that does not need the reactor or any other
+asynchronous process in order to fire.
+
+Normal application code can't know when a Deferred is going to fire, because
+that is generally left up to the reactor. Well-written unit tests provide fake
```
rbtcollins: comment only: I realise that this is a true statement, but it sounds a little judgy.

testtools/_deferredmatchers.py
```diff
@@ -0,0 +1,188 @@
...
+    As will this:
+
+    >>> assert_that(defer.fail(RuntimeError('foo')), no_result())
+    Traceback (most recent call last):
+      ...
+      File "testtools/assertions.py", line 22, in assert_that
+        raise MismatchError(matchee, matcher, mismatch, verbose)
+    testtools.matchers._impl.MismatchError: No result expected on <Deferred at ... current result: <twisted.python.failure.Failure <type 'exceptions.RuntimeError'>>>, found <twisted.python.failure.Failure <type 'exceptions.RuntimeError'>> instead
+    """
+    return _NoResult()
```
rbtcollins: Yeah, a constant IMO

testtools/_deferredmatchers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Use this instead of
+    :py:meth:`twisted.trial.unittest.SynchronousTestCase.failureResultOf`.
+
+    :param matcher: A matcher to match against the result of a failing
+        :class:`~twisted.internet.defer.Deferred`.
+    :return: A matcher that can be applied to a synchronous
+        :class:`~twisted.internet.defer.Deferred`.
+    """
+    return _Failed(matcher)
```
rbtcollins: This is equivalent to failed = _Failed - I don't understand why the indirection. Same as above - apply this question there ;)

rbtcollins:
I think no_result should be made a constant. Agree?

Either that or just not exist, since its just a trivial lambda.

    What do we think of the names?

Purple.

    Where should I export them? I'm thinking of creating a new module testtools.twisted and exporting them there (along with DeferredRunTest and kin)

That would be ok with me, just don't call it twisted, because name shadowing bad mkay.

    Should failed take exception types? If so, can we defer to another branch?

No, they can be tunnelled through via the matcher.

doc/conf.py
jml: Renamed to 'py2' and added 'py3'.

testtools/_deferred.py
jml: Amended as per example.

testtools/_deferred.py
jml: Changed to:

>  As a rule, this function should not be used when operating with asynchronous Deferreds (i.e. for normal use of Deferreds in application code). In those cases, it is better to add callbacks and errbacks as needed.

testtools/_deferred.py
jml: Only if code is broken. Note that I only moved this function.

testtools/_deferred.py
jml: Weird is subjective. Closing paren aligning with opening statement is much more readable to me.

Amended.

testtools/_deferredmatchers.py
jml: Fair. Changed to:

> Unit tests can (and should!) provide …

testtools/_deferredmatchers.py
jml: To prevent subclassing so that we're not supporting subclassing as an API in the future. We could do this from documentation, but this does better on the Rusty scale.

jml:
Thanks for the review!

* Created `twistedsupport` module that exports these matchers plus the "public" stuff from deferredruntest (it's not 100% clear what's public, so I've settled for everything that was referred to in the documentation)
* Renamed `no_result()` to `NO_RESULT`
* Will ask colleagues for a quick double-check of the names

NEWS
```diff
@@ -27,6 +27,13 @@ Improvements
 * A failing ``expectThat`` now fails tests run with
   ``AsynchronousDeferredRunTest``.  (Jonathan Lange, #1532452)
 
+* New ``testtools.twistedsupport`` module that collects all of our Twisted
+  support code in one place, including that currently available under
+  ``testtools.deferredruntest``.  (Jonathan Lange)
+
+* New matchers for testing ``Deferred`` code: ``failed``, ``successful``, and
+  ``NO_RESULT``.  (Jonathan Lange, Tristan Seligmann, #1369134)
```
exarkun: From the outside, it's seems a little bit weird to have `failed` and `successful` but then `NO_RESULT`.  I don't know what the model is that makes that the right choice.

Also, less surprising but slightly annoying (and certainly a nit if there is such a thing), I would say that a Deferred "was successful" but I wouldn't say it "was failed".  The lack of parallelism irks slightly.  Using `failed` and `succeeded` would address this - both "the Deferred failed" and "the Deferred succeeded".

doc/twisted-support.rst
```diff
@@ -1,13 +1,51 @@
...
+:py:class:`~twisted.internet.defer.Deferred`\s.
+
+A "synchronous" :py:class:`~twisted.internet.defer.Deferred` is one that does
+not need the reactor or any other asynchronous process in order to fire.
+
+Normal application code can't know when a
+:py:class:`~twisted.internet.defer.Deferred` is going to fire, because that is
+generally left up to the reactor. Well-written unit tests provide fake
+reactors, or don't use the reactor at all, so that
+:py:class:`~twisted.internet.defer.Deferred`\s fire synchronously.
```
exarkun: I guess you probably checked how this renders but I thought you had to do "...Deferred\`\ s"` to get this to work. (note the space after the backslash).

testtools/_deferred.py
```diff
@@ -0,0 +1,112 @@
...
+
+from testtools.content import TracebackContent
+
+
+class DeferredNotFired(Exception):
+    """Raised when we extract a result from a Deferred that's not fired yet."""
+
+    def __init__(self, deferred):
+        msg = "%r has not fired yet." % (deferred,)
+        super(DeferredNotFired, self).__init__(msg)
```
exarkun: Intentionally not keeping the Deferred as an attribute of the exception?  The repr of a Deferred is usually nearly useless in my experience.  Maybe knowing its id() is useful for correlating it with another occurrence... apart from that it's just noise, though.

testtools/_deferredmatchers.py
```diff
@@ -0,0 +1,184 @@
...
+            on_no_result=lambda _: None,
+        )
+
+
+NO_RESULT = _NoResult()
+"""Match a Deferred that has not yet fired.
+
+For example, this will pass::
+
+    assert_that(defer.Deferred(), NO_RESULT)
```
exarkun: Okay I guess the property of the model that suggests the naming divergence between `NO_RESULT` and the others is that `NO_RESULT` is a singleton matcher, not a function of one argument that returns a matcher.

NEWS
exarkun: And the prevailing matcher naming convention suggests some other names, I think - `fails` and `succeeds`.  Also ... `continues` (or something) for `NO_RESULT` but that's probably 98% too cute.

exarkun:
Thanks.  Some comments inline.  Overall I think this will provide useful functionality as-is and my feedback is mostly around providing a little more polish.

NEWS
jml: I like `failed` and `succeeded`. Will change.

As for `NO_RESULT` (an obvious constant) vs `no_result` (symmetric, guessable name), I'm very ambivalent.

doc/twisted-support.rst
jml: Yeah, I did check—I spent large chunks of the weekend learning about Sphinx. I'm pretty sure it works without the space, but will double check before landing, to make assurance double sure.

testtools/_deferred.py
jml: Why would it help to have it as an attribute? For interactive debugging?

NEWS
exarkun: A solution for the constant / symmetric tension would be to make it not a constant, even though it *can* be a constant.  `assertThat(foo, no_result())` (and then I'm tempted to re-raise the idea of naming it `has_no_result()`).

testtools/_deferred.py
exarkun: The main thing that I think it would help is testing the implementation... The assertion about an attribute referencing the right object is more direct than an assertion about a string containing the right bytes.  That's not a very compelling user-facing scenario - but sometimes when improving things so they're easier to test, unanticipated improvements are yielded for end-users as well.

Definitely no really strong feelings here, just a vague hunch keeping things structured when possible is better than tossing strings around.

jml:
> Created twistedsupport module that exports these matchers plus the "public" stuff from deferredruntest (it's not 100% clear what's public, so I've settled for everything that was referred to in the documentation)

I have [a branch](https://github.com/jml/testtools/tree/twistedsupport-package) on top of this one that turns `twistedsupport` into a package and:
* moves all of the Twisted code into it
* makes a `testtools.tests.twistedsupport` package (parallel to `testtools.tests.matchers`)
* creates a `testtools.deferredruntest` module for backwards compatibility (currently no deprecation warnings, but could add them easily enough)

Would not be averse to doing it in the same commit.

jml:
Merging, since all review comments addressed.

PR #194, opened 2016-01-09
expectThat works under AsynchronousDeferredRunTest

jml:
`AsynchronousDeferredRunTest` has an unfortunate bug: if `expectThat` is called and fails, the test passes. See https://bugs.launchpad.net/testtools/+bug/1532452.

This is a critical bug.

There's a deeper bug in our testing strategy for `RunTest`. However, that will be vastly easier to address after #165 lands (and maybe after #181, which will provide us with an interface library), so I'm not tackling it here.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/194)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -24,6 +24,10 @@ Improvements
 * Tests that customize ``skipException`` no longer get tracebacks for skipped
   tests.  (Jonathan Lange)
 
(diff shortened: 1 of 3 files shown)


rbtcollins:
LGTM

PR #196, opened 2016-01-22
Handle edge case in reactor running

jml:
This was a tricky one to diagnose, but hopefully the test is straightforward and well-commented enough to explain what's going on.

I'd like to follow up with some sort of logging thing, but that's surprisingly fraught (which logger? what levels? how passed from `RunTest`? If Twisted logger, how does it interact with the users own test logs? With our tests for handling users logs?), and would also clash with #171 and #172, which are much more important, and really should land before anything of the sort.

I'd also like to fix `ADRT` to not raise `KeyboardInterrupt`, but to instead log an error and stop the run. *That's* blocked on Trial not supporting `stop`/`shouldStop` and also by #171 and #172.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/196)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -38,6 +38,11 @@ Improvements
   between runs. In particular, testtools tests can now be run with
   ``trial -u``.  (Jonathan Lange, #1517879)
 
+* Fixed bug where if an asynchronous ``Deferred`` test times out but the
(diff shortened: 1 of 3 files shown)


jml:
Review timeout.

PR #197, opened 2016-01-25
Fix Travis builds for pypy3 and Python 3.2

jml:
Recent releases of pip 8.0 and setuptools 19.4.1 broke things for Python 3.2 and for pypy3. This PR fixes the Travis build by anchoring those dependencies accordingly.

The next release of testtools is to be the last that supports 3.2, so this is a very temporary measure.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/197)


jml:
Merging without review in order to get the next release out.

PR #201, opened 2016-01-27
Use `skipTest` instead of `skip`

jml:
`skip` is deprecated, so updates tests to use `skipTest`.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/201)


jml:
Review timeout. Merged as b3394b2a5c7c8ba2986b1d93518655fd067217dc

PR #202, opened 2016-01-27
Move all Twisted support code to twistedsupport package

jml:
Intent is purely to move code around without causing behavior change, so that all the Twisted stuff is in one spot.

* Move all Twisted support code to private modules in `testtools.twistedsupport` package
* Export all public code from `testtools.twistedsupport` package
* Continue to export public code from old locations (e.g. so `testtools.deferredruntest.AsynchronousDeferredRunTest` still works)
* Move all tests for Twisted support code to `testtools.tests.twistedsupport`, to parallel matchers
* Move `NeedsTwistedTestCase` to its own helper module, rather than importing from `test_spinner` as before

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/202)

PR #204, opened 2016-01-28
Disable Twisted debugging in some tests

jml:
Some of our tests assume that debugging is not enabled in Twisted. This makes that assumption explicit by disabling debugging for the duration of the test.

It does this by:
* changing the spinner to use a custom MonkeyPatch fixture, rather than testtools monkey patch
* making the fixture explicitly set debugging bools to a value (rather than hardcoding true or false)
* using a fixture always, but only the identity fixture (`Fixture`) when debugging is not set
* updating the relevant tests to use `DebugTwisted(False)`

Testing strategy has been to run tests with `trial --debug-stacktraces`. I don't think we need explicit tests for the debugging behavior. There are other issues preventing trial from running in full (due to Trial's weird reporter API), but that's a separate PR for another time.

Will conflict with #171, which also adds fixtures as a dependency, and (in spirit) with #202, which moves all Twisted code to its own package.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/204)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,9 @@ Changes
 * Python 2.6 and 3.2 are no longer supported. If you want to use either of
   these versions of Python, use testtools 1.9.0. (Jonathan Lange)
 
+* Make ``fixtures`` a real dependency, not just a test dependency.
+  (Jonathan Lange)
+
 
 1.9.0
 ~~~~~
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 pbr>=0.11
 extras
+fixtures
 pyrsistent
 # 'mimeparse' has not been uploaded by the maintainer with Python3 compat
 # but someone kindly uploaded a fixed version as 'python-mimeparse'.
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,6 @@ classifier =
 
 [extras]
 test =
-  fixtures
   testscenarios
   unittest2>=1.1.0
 
--- a/testtools/_deferreddebug.py
+++ b/testtools/_deferreddebug.py
@@ -1,0 +1,23 @@
+# Copyright (c) testtools developers. See LICENSE for details.
+#
+# TODO: Move this to testtools.twistedsupport. See testing-cabal/testtools#202.
+
+from fixtures import Fixture, MonkeyPatch
+
+
+class DebugTwisted(Fixture):
+    """
+    Set debug options for Twisted.
+    """
+
+    def __init__(self, debug=True):
+        super(DebugTwisted, self).__init__()
+        self._debug_setting = debug
+
+    def _setUp(self):
+        self.useFixture(
+            MonkeyPatch('twisted.internet.defer.Deferred.debug',
+                        self._debug_setting))
+        self.useFixture(
+            MonkeyPatch('twisted.internet.base.DelayedCall.debug',
+                        self._debug_setting))
--- a/testtools/_spinner.py
+++ b/testtools/_spinner.py
@@ -16,12 +16,12 @@ __all__ = [
     'trap_unhandled_errors',
     ]
 
+from fixtures import Fixture
 import signal
 
-from testtools.monkey import MonkeyPatcher
+from testtools._deferreddebug import DebugTwisted
 
 from twisted.internet import defer
-from twisted.internet.base import DelayedCall
 from twisted.internet.interfaces import IReactorThreads
 from twisted.python.failure import Failure
 from twisted.python.util import mergeFunctionMetadata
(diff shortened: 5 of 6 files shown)


rbtcollins:
If this can be landed now, lets do that and avoid juggling conflicts

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[testtools/_deferreddebug.py, line 7 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_SctwAcu6rn4mBCp:-K9B_SctwAcu6rn4mBCq:875228998) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L7)):*
No __all__ ?

---

*[testtools/_deferreddebug.py, line 11 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_VptFJHMIrpw0Xv7:-K9B_VptFJHMIrpw0Xv8:1317346586) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L11)):*
PEP 297 (I think?) Specifically shows """ Foo bar baz""" for one liners. I value vertical space a lot -  I realise its not a functional thing, but I'd like to keep this consistent, and I'm pretty sure we've done genuine one liners for everything in the code base...

---

*[testtools/_deferreddebug.py, line 17 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9BMSHhTdkxKLA10TIJ:-K9BMSHhTdkxKLA10TIK:-704273898) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L17)):*
Since this is using _setUp, the dependency on fixtures should be versioned with minimum appropiate to the introduction of that API.

---

*[testtools/_deferreddebug.py, line 23 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_yWWCf_TTY_BniwU:-K9B_yWWCf_TTY_BniwV:607288946) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L23)):*
I think this is fine as-is. It occurred to me reading it that you could instead say
```
def __init__(....)
    super
    self._deferred_debug = MoneyPatch('twisted.internet.defer.Deferred.debug', debug)
    self._delayedcall_debug = MonkeyPatch('twisted.internet.base.DelayedCall.debug', debug)

def _setUp(self):
    self.useFixture(self._deferred_debug)
    self.useFixture(self._delayedcall_debug)
```

Which seems a little clearer to me, but YMMW

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-:-K9BaQguTaUFE-51mtCN:-196284059)*

jml:
Thanks for the review. I've updated the docstring, and will merge. 

---

Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[testtools/_deferreddebug.py, line 7 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_SctwAcu6rn4mBCp:-K9GpcDGfHD0nRo34BSR:1186275135) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L7)):*
No. I don't think it's quite so important for internal code. Can add one later if you feel strongly about it, or we have a coherent standard written somewhere.

---

*[testtools/_deferreddebug.py, line 11 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_VptFJHMIrpw0Xv7:-K9Gp9rCOlhTuv8PR1If:2066690876) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L11)):*
Yeah, we are. flocker follows this (Twisted-style) approach to docstrings and it creeped over.

---

*[testtools/_deferreddebug.py, line 17 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9BMSHhTdkxKLA10TIJ:-K9GppuZEnMGzf_AuliV:-499146009) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L17)):*
Done.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-:-K9Gq2ap1HH7EuV2lv4l:-218260291)*

jml:
Review status: 0 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[testtools/_deferreddebug.py, line 23 \[r1\]](https://reviewable.io:443/reviews/testing-cabal/testtools/204#-K9B_yWWCf_TTY_BniwU:-K9GrPmKTNb8CsgpOMWE:752462935) ([raw file](https://github.com/testing-cabal/testtools/blob/4fa4e77589aa3372588932d60ea9e5fb0512fe10/testtools/_deferreddebug.py#L23)):*
The way I wanted to write it was:

```
def debug_twisted(debug):
    return CompoundFixture([
        MonkeyPatch('twisted....', debug),
        MonkeyPatch('twisted....', debug),
    ])
```

It was this second use-case which provoked me to testing-cabal/fixtures#22

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/testing-cabal/testtools/204)*

jml:
Merged at  91c195c

PR #178, opened 2015-12-10
[WIP] Explicit interfaces

jml:
* Uses zope.interface
* Not yet complete
* Documenting what we've got, even if I would like to change it

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/178)

PR #211, opened 2016-02-05
Explicit interfaces for matchers and mismatches

jml:
* Adds `IMatcher` and `IMismatch`
* Updates their documentation as well as that in `assert_that`,
  `assertThat`, and `expectThat` to refer to these
* Changes `MismatchDecorator` to explicitly subclass `IMismatch`

Another take on #181. This time, just using Python.

While this patch is not a huge improvement in and of itself, I want to start doing this (explicit abstract types) for the rest of testtools. I think cumulatively it will make a great improvement to the code base. (Already the work on #178 has helped me clarify much). Matchers are an easy place to start because they are so close to this already.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/211)

### DIFF
--- a/testtools/assertions.py
+++ b/testtools/assertions.py
@@ -14,7 +14,6 @@ def assert_that(matchee, matcher, message='', verbose=False):
     :param matchee: An object to match with ``matcher``.
     :param IMatcher matcher: An object meeting the testtools.Matcher protocol.
     :raises MismatchError: When ``matcher`` does not match ``matchee``.
-
     """
     matcher = Annotate.if_message(message, matcher)
     mismatch = matcher.match(matchee)
--- a/testtools/assertions.py
+++ b/testtools/assertions.py
@@ -7,13 +7,14 @@ from testtools.matchers import (
 def assert_that(matchee, matcher, message='', verbose=False):
     """Assert that matchee is matched by matcher.
 
-    This should only be used when you need to use a function based
-    matcher, assertThat in Testtools.Testcase is prefered and has more
-    features
+    This should only be used when you need to use a function-based matcher,
+    ``assertThat`` in :py:class:`testtools.TestCase` is prefered and has more
+    features.
 
-    :param matchee: An object to match with matcher.
-    :param matcher: An object meeting the testtools.Matcher protocol.
-    :raises MismatchError: When matcher does not match thing.
+    :param matchee: An object to match with ``matcher``.
+    :param IMatcher matcher: An object meeting the testtools.Matcher protocol.
+    :raises MismatchError: When ``matcher`` does not match ``matchee``.
+
     """
     matcher = Annotate.if_message(message, matcher)
     mismatch = matcher.match(matchee)
--- a/testtools/matchers/_impl.py
+++ b/testtools/matchers/_impl.py
@@ -122,14 +122,12 @@ class Mismatch(IMismatch):
         self._details = details
 
     def describe(self):
-        """Describe the mismatch."""
         try:
             return self._description
         except AttributeError:
             raise NotImplementedError(self.describe)
 
     def get_details(self):
-        """Get extra details about the mismatch."""
         return getattr(self, '_details', {})
 
     def __repr__(self):
--- a/testtools/matchers/_impl.py
+++ b/testtools/matchers/_impl.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2009-2012 testtools developers. See LICENSE for details.
+# Copyright (c) testtools developers. See LICENSE for details.
 
 """Matchers, a way to express complex assertions outside the testcase.
 
@@ -11,6 +11,7 @@ $ python -c 'import testtools.matchers; print testtools.matchers.__all__'
 """
 
 __all__ = [
+    'IMatcher',
     'Matcher',
     'Mismatch',
     'MismatchDecorator',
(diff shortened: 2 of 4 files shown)


testtools/matchers/_impl.py
```diff
@@ -53,7 +62,49 @@ def __str__(self):
...
+Matcher = IMatcher
+
+
+class IMismatch(object):
+    """A mismatch detected by an IMatcher.
+
+    Describes the protocol that mismatches are required to implement.
+    """
+
+    # XXX: jml would like to extend this interface to include the thing that
```
thomir: FWIW, I'm +1 on this - I frequently find myself writing a test that fails, then adding the matchee as the third parameter to the assertion so it gets printed out, then fixing the test and removing that third parameter. Annoying. I'm not sure if that should go in the matchers themselves, or in (assert|expect)That.

testtools/matchers/_impl.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2009-2012 testtools developers. See LICENSE for details.
+# Copyright (c) testtools developers. See LICENSE for details.
```
thomir: Aren't copyright years worth keeping?

thomir:
A few comments above. Probably worth getting @rbtcollins opinion as well, but I'm +1 anyway.

testtools/matchers/_impl.py
jml: Ah, interesting. I mostly want it for composition, e.g. to help with `Not`. ISTR wanting it for https://github.com/ClusterHQ/flocker/blob/master/flocker/testtools/matchers.py but looking at it, I can't see what was essential.

Whether we or not we extend the interface this way, I want second opinions on whether it should be done in this patch / this release.

testtools/matchers/_impl.py
jml: Opinion differs. Fixed.

jml:
Thanks. I'll wait for @rbtcollins, because he had opinions about #181.

testtools/assertions.py
```diff
@@ -7,13 +7,14 @@
...
-    This should only be used when you need to use a function based
-    matcher, assertThat in Testtools.Testcase is prefered and has more
-    features
+    This should only be used when you need to use a function-based matcher,
+    ``assertThat`` in :py:class:`testtools.TestCase` is prefered and has more
+    features.
+
+    :param matchee: An object to match with ``matcher``.
+    :param IMatcher matcher: An object meeting the testtools.Matcher protocol.
+    :raises MismatchError: When ``matcher`` does not match ``matchee``.
```
rbtcollins: spurious VWS

testtools/matchers/_impl.py
```diff
@@ -53,7 +62,49 @@ def __str__(self):
...
+
+    def describe(self):
+        """Describe the mismatch.
+
+        :return: Either a human-readable string or something that can be cast
+            to a string. On Python 2, this should be either ``unicode`` or a
+            plain ASCII ``str``, and care should be taken to escape control
+            characters.
+        """
+        raise NotImplementedError(self.describe)
```
rbtcollins: I'm not sure, but I think we want this to actually be a string, always.

testtools/matchers/_impl.py
```diff
@@ -71,33 +122,14 @@ def __init__(self, description=None, details=None):
         self._details = details
 
     def describe(self):
-        """Describe the mismatch.
-
-        This should be either a human-readable string or castable to a string.
-        In particular, is should either be plain ascii or unicode on Python 2,
-        and care should be taken to escape control characters.
-        """
+        """Describe the mismatch."""
```
rbtcollins: I'd rather have no docstring here rather than an abreviated one, perhaps a decorator to consume the base class docstring?

testtools/matchers/_impl.py
```diff
@@ -71,33 +122,14 @@ def __init__(self, description=None, details=None):
...
-        The testtools assertion assertThat will query get_details and attach
-        all its values to the test, permitting them to be reported in whatever
-        manner the test environment chooses.
-
-        :return: a dict mapping names to Content objects. name is a string to
-            name the detail, and the Content object is the detail to add
-            to the result. For more information see the API to which items from
-            this dict are passed testtools.TestCase.addDetail.
-        """
+        """Get extra details about the mismatch."""
```
rbtcollins: Ditto.

rbtcollins:
This looks much saner to me.

testtools/matchers/_impl.py
jml: What do you mean by "string"?

testtools/matchers/_impl.py
jml: @rbtcollins Any thoughts on this?

testtools/assertions.py
jml: Fixed

testtools/matchers/_impl.py
jml: Updated to have no docstring. Decorator can come later.

testtools/matchers/_impl.py
jml: As above.

jml:
Updated based on review comments. Need clarification on a couple of points before I land.

jml:
Don't want this on my Github dashboard. Others are free to re-submit if they'd like.

testtools/matchers/_impl.py
freeekanayaka: @jml Just for extra clarity, does "include the thing that" mean "the matchee argument passed to match()"?

From a very quick look at the code it seems that implementations of IMatcher typically pass or could pass the matchee object to the constructor of an IMismatch, in order to have describe() take it into account.

Why do we need to pass it to describe()?

An example of the "composition" use case you allude to would be appreciated.

jml: >  does "include the thing that" mean "the matchee argument passed to match()"

Yes.

> Why do we need to pass it to describe()?

Did I propose this?

> An example of the "composition" use case you allude to would be appreciated.

Sorry, the intervening 13 months have somehow scrubbed it from my memory. However, looking at the [file linked to in my comment](https://github.com/ClusterHQ/flocker/blob/master/flocker/testtools/matchers.py#L176) provides just such an example. You'll have to figure out what it does, I'm afraid.

freeekanayaka:
Jonathan Lange <notifications@github.com> writes:

> jml commented on this pull request.
>
>
>
>> @@ -53,7 +62,49 @@ def __str__(self):
>          raise NotImplementedError(self.__str__)
>  
>  
> -class Mismatch(object):
> +Matcher = IMatcher
> +
> +
> +class IMismatch(object):
> +    """A mismatch detected by an IMatcher.
> +
> +    Describes the protocol that mismatches are required to implement.
> +    """
> +
> +    # XXX: jml would like to extend this interface to include the thing that
>
>>  does "include the thing that" mean "the matchee argument passed to match()"
>
> Yes.
>
>> Why do we need to pass it to describe()?
>
> Did I propose this?

I might have misunderstood. However what does

```python
# XXX: jml would like to extend this interface to include the thing that
# failed to match.
```

refer to then? If you recall. IOW how would such an extension look like.

jml:
A public `mismatched` attribute, as exemplified in the linked `flocker/matchers.py`.

Regrettably, this is my last update to this PR. I don't use Python at home and I don't use testtools at work, so I'm not going to make time to pursue this further.

freeekanayaka:
@jml thanks for the input

PR #213, opened 2016-02-08
Better handling of KeyboardInterrupt

jml:
When we get a suspected KeyboardInterrupt (no result from Deferred),
explicitly stop the result and report the error as if it were a user
error.

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/213)


jml:
Merged in cd286b23f23c2d79b60e1bb3686d7a17507595b8

PR #220, opened 2016-03-03
Revert usage of pyrsistent and _TestRecord

### LINKED ISSUE #219: Performance regression with testtools 1.9
I realized recently that testr got super slow.

On my repository:

> time testr last
Ran 5826 tests in 483.814s
PASSED (id=0, skips=5)

real	1m41.814s
user	1m41.197s
sys	0m0.082s

I tried a bunch of things, and finally: 

> pip install testtools==1.8.0

> time testr last
Ran 5826 tests in 483.814s
PASSED (id=0, skips=5)

real	0m20.736s
user	0m20.520s
sys	0m0.051s

So there is a giant performance regression here. I'll try to track it down, but I wanted to file the issue first.

therve:
The _TestRecord class is pretty big performance hit, as it adds a fair
amount of indirection for each test execution. This reverts its creation
and remove the dependency on pyrsistent. 

Close #219

[Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/220)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,6 @@
 pbr>=0.11
 extras
 fixtures>=1.3.0
-pyrsistent
 # 'mimeparse' has not been uploaded by the maintainer with Python3 compat
 # but someone kindly uploaded a fixed version as 'python-mimeparse'.
 python-mimeparse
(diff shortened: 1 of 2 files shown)


rbtcollins:
I'll apply this in a few days so that jml and thomir get a chance to review for other ways to address it, but yeah, looks straight forward enough to me; sad tho.

jml:
I'd have done this by changing the pyrsistent class to a regular one and keeping the rest: I think it's a lot better than an implicit dict. However, I don't think that strongly enough to object, so +0 from me.

rbtcollins:
@therve could you perhaps test the performance with a regular class?

therve:
I tried with https://github.com/testing-cabal/testtools/pull/221.

It's still a bit slower, but more in the 10% range rather than the 200% range.

PR #224, opened 2016-03-12
Implement matchers for warnings.

### LINKED ISSUE #223: Matcher for warnings
Capturing warnings and then asserting stuff about them is necessary work when deprecating things, for example, but it's not really much fun to write the boilerplate oneself.

jonathanj:
Introduce a matcher for capturing `warning.WarningMessage`s when
executing a function, a function for matching `warning.WarningMessage`
structure and a function for matching deprecation warnings.

Fixes #223.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/224)

### DIFF
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -491,7 +491,7 @@ Most of the time, you will want to uses `The raises helper`_ instead.
 warning_message
 ~~~~~~~~~~~~~~~
 
-Match against various attributes–category, message and filename to name a few—of
+Match against various attributes (category, message and filename to name a few) of
 a `warning.WarningMessage`, which can be captured by `Warnings`_.
 
 
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -392,6 +392,25 @@ This is actually a convenience function that combines two other matchers:
 Raises_ and MatchesException_.
 
 
+The is_deprecated helper
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Matches if a callable produces a warning whose message matches the specified
+matcher. For example::
+
+  def old_func(x):
+      warnings.warn('old_func is deprecated use new_func instead')
+      return new_func(x)
+
+  def test_warning_example(self):
+      self.assertThat(
+        lambda: old_func(42),
+        is_deprecated(Contains('old_func is deprecated')))
+
+This is just a convenience function that combines Warnings_ and
+`warning_message`_.
+
+
 DocTestMatches
 ~~~~~~~~~~~~~~
 
@@ -467,6 +486,13 @@ example::
       self.assertThat(exc_info, MatchesException(RuntimeError('bar')))
 
 Most of the time, you will want to uses `The raises helper`_ instead.
+
+
+warning_message
+~~~~~~~~~~~~~~~
+
+Match against various attributes–category, message and filename to name a few—of
+a `warning.WarningMessage`, which can be captured by `Warnings`_.
 
 
 NotEquals
@@ -909,6 +935,28 @@ Although note that this could also be written as::
       self.assertThat(lambda: 1/0, raises(ZeroDivisionError))
 
 See also MatchesException_ and `the raises helper`_
+
+
+Warnings
+~~~~~~~~
+
+Captures all warnings produced by a callable as a list of
+`warning.WarningMessage` and matches against it. For example, if you want to
+assert that a callable emits exactly one warning::
+
+  def soon_to_be_old_func():
+      warnings.warn('soon_to_be_old_func will be deprecated next version',
+                    PendingDeprecationWarning, 2)
+
+  def test_warnings_example(self):
+      self.assertThat(
+          soon_to_be_old_func,
+          Warnings(HasLength(1)))
+
+Deprecating something and making an assertion about the deprecation message is a
+very common test with which `the is_deprecated helper`_ can assist. For making
+more specific matches against warnings `warning_message`_ can construct a
+matcher that can be combined with `Warnings`_.
 
 
 Writing your own matchers
--- a/doc/for-test-authors.rst
+++ b/doc/for-test-authors.rst
@@ -392,7 +392,7 @@ This is actually a convenience function that combines two other matchers:
 Raises_ and MatchesException_.
 
 
-The is_deprecated helper
+The IsDeprecated helper
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 Matches if a callable produces a warning whose message matches the specified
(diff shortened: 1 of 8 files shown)


doc/for-test-authors.rst
```diff
@@ -392,6 +392,25 @@ This is actually a convenience function that combines two other matchers:
 Raises_ and MatchesException_.
 
 
+The is_deprecated helper
```
thomir: I suggest changing the name - 'is_deprecated' is too specific and implies something about the warning message that we don't know.

doc/for-test-authors.rst
```diff
@@ -469,6 +488,13 @@ example::
 Most of the time, you will want to uses `The raises helper`_ instead.
 
 
+warning_message
+~~~~~~~~~~~~~~~
+
+Match against various attributes–category, message and filename to name a few—of
```
thomir: Suggest using parenthesis instead of hyphens, as I think it makes it easier to read:

> Match against various attributes (category, message and filename to name a few) of...

testtools/matchers/_warnings.py
```diff
@@ -0,0 +1,97 @@
...
+def warning_message(category_type, message=None, filename=None, lineno=None,
+                    line=None):
+    """
+    Create a matcher that will match `warnings.WarningMessage`\s.
+
+    :param category_type: A warning type.
+    :param message_matcher: Match the warning message against this.
+    :param filename_matcher: Match the warning filename against this.
+    :param lineno_matcher: Match the warning line number against this.
+    :param line_matcher: Match the warning line of code against this.
```
thomir: I think it may be confusing that some of these parameters are types, and some are matchers. There are a few things we could do in the docstring to make this more clear:

 * An example would help people understand how this function is intended to be used.
 * The individual :param: lines could be clearer. For example:

 > :param message_matcher: A matcher object that will be evaluated against the message category.

testtools/matchers/_warnings.py
```diff
@@ -0,0 +1,97 @@
...
+from ._higherorder import (
+    AfterPreprocessing,
+    Annotate,
+    MatchesAll,
+    Not,
+    )
+from ._impl import Mismatch
+
+
+def warning_message(category_type, message=None, filename=None, lineno=None,
```
thomir: @rbtcollins @jml : did we ever settle on a standard naming convention for matcher types? Having some matchers as functions and some as classes is a little irksome. I'm not sure what we should do at this point:

 * Follow pep8, and force users to import a mix of `function_name` and `ClassName` matchers?
 * Make functions follow `ClassName` convention (FWIW, we do this in `testtools.content` a fair bit).

Thoughts?

testtools/matchers/_warnings.py
```diff
@@ -0,0 +1,97 @@
...
+            "Warning's line number does not match",
+            lineno_matcher),
+        line=Annotate(
+            "Warning's source line does not match",
+            line_matcher))
+
+
+class Warnings(object):
+    """
+    Match if the matchee produces deprecation warnings.
```
thomir: I suggest removing the word 'deprecation' - the warning could be about anything, not just deprecated code, right?

testtools/matchers/_warnings.py
```diff
@@ -0,0 +1,97 @@
...
+        matchee. If no warnings_matcher is supplied then the simple fact that
+        at least one warning is emitted is considered enough to match on.
+        """
+        self.warnings_matcher = warnings_matcher
+
+    def match(self, matchee):
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter('always')
+            matchee()
+            if not w:
```
thomir: Can I use this to assert that the matchee produces no warnings?

self.assertThat(my_fn, Warnings(HasLength(0)))

?

Is that a use case that we want to support?

thomir:
Hi,

This looks good - I've made a few notes, and I'm sure @rbtcollins or @jml  will want to take a look as well.

Don't forget to add an item to NEWS, and add your name to the authors list if it's not already there. 

Thanks,

testtools/matchers/_warnings.py
jonathanj: @thomir After a discussion with @jml yesterday, my understanding is that a large percentage of matcher implementations should just be a function (the matching logic) instead of deriving a new type. Following `ClassName` convention for functions might make the API _appear_ more consistent but it's a bit misleading because it has none of the applications of a function, my 2c would be keeping the conventions for functions and classes separate. (`raises` and `Raises` are possible examples.)

testtools/matchers/_warnings.py
jonathanj: I think your matcher would work for that case, perhaps a helper function `no_warnings` (?) might make for more intuitive reading?

testtools/matchers/_warnings.py
jonathanj: Indeed! This was a remnant from a previous implementation.

testtools/matchers/_warnings.py
jonathanj: Agreed. I've updated the docstring with your suggestions.

doc/for-test-authors.rst
jonathanj: Agreed. Done.

jonathanj:
Thanks for the prompt review, @thomir! I think I've made changes to address all of your review commentary and replied to the discussion points.

jml:
> @rbtcollins @jml : did we ever settle on a standard naming convention for matcher types? Having some matchers as functions and some as classes is a little irksome. I'm not sure what we should do at this point:
>
> Follow pep8, and force users to import a mix of function_name and ClassName matchers?
> Make functions follow ClassName convention (FWIW, we do this in testtools.content a fair bit).

I'd rather follow pep8, but with the `Always` patch we agreed that the convention would be to use `CamelCase` even if we're exposing a function.

jml:
@thomir I trust your judgment on this. No need for a review from me.

testtools/matchers/_warnings.py
thomir: IMO, `raises` and `Raises` are an anti-pattern. As a testtools user, I can never remember which one does what (although that might have more to do with having two objects with the same name and different case). As @jml commented on the MP, We recently decided to export functions as CamelCase names for the sake of consistency. That decision isn't perfect, but I think in this case there's no good answer, and 'CamelCase'ing everything is probably the best of a bunch of bad options.

testtools/matchers/_warnings.py
thomir: So, just from reading the code, I don't think it will, since you're always checking `if not w`, which will always return a Mismatch. Further, `self.warnings_matcher` is only evaluated if `if not w` evaluates to false. 

If we want to support this use case (and I don't see any reason not to), I think it may be worth adding a test case for it. I am, after all, just reading the code without actually running it (I'm feeling lazy today).

jonathanj: Oops, you're quite right! I've made this use case possible and added a test for it.

testtools/matchers/_warnings.py
jonathanj: Okay, renamed `is_deprecated` and `warning_message` to `IsDeprecated` and `WarningMessage` respectively.

jonathanj:
@thomir I believe I've addressed all previous review comments.

thomir:
Thanks!

Merged @ 8a69c2c536f9f86931c0aafc4153268ec17305c7.

PR #232, opened 2016-05-18
Unbreak fixtures test suite / use of gather_details

rbtcollins:
There is a circular import loop between testtools and fixtures,
which is fine, but we need to define gather_details before importing
fixtures.

Change-Id: I9a59968c4e2ac2c29d679d7f5c474696c898671e

---
This change is [Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/232)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 pbr>=0.11
-extras
+extras>=1.0.0
 fixtures>=1.3.0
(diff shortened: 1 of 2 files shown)


rbtcollins:
This is incomplete, there's also a bug in extras; I'm thinking to fix that and make this also bump our deps

thomir:
LGTM

rbtcollins:
APplied.

Issue #239, opened 2016-09-16
Allow `expectThat` to return True/False if the condition passed or failed.


blakev:
Because `expectThat` does a differed assertion, it would be nice to be able to know while the `TestCase` is running whether or not the condition passed for advanced logging or logic.

Our use case is using `expectThat` matchers in long running tests that need to log the success AND failures of those checks, not just the ones that failed at the end of the run (returned in `getDetails()`).

thomir:
HI @blakev,

I don't understand the need for this change. perhaps you could show an example test where the existing functionality makes things awkward and this feature would help?

Generally speaking, I think adding flow control (if/else/etc) statements to a test is a bad idea: It becomes harder to reason about what was tested by looking at the code. If you find yourself needing to do this, there's almost certainly a better way - perhaps by splitting the tests into two or more tests that test each path separately.

Cheers,

blakev:
`expectThat` isn't actually asserting anything while the test is running, instead they're collected at the end of the method and reported all at the same time. Because code will continue executing after it's been called you can only get the details of the `expectThat` at the end. The test has no conditional logic, instead the logging/reporting metrics need to know which expects passed and which failed.

The updated return statement doesn't break anything, because you wouldn't be calling the method to expect a `None` no matter what.

thomir:
Hi @blakev,

You are correct about the return value - that's what happens when you review code before your first coffee in the morning, sorry!

I still don't understand the rationale for wanting this. You say "The test has no conditional logic, instead the logging/reporting metrics need to know which expects passed and which failed." - which reporting mechanisms, exactly? Perhaps you could show us what you're ultimately trying to achieve here? Some example code would help, I think.

blakev:
Consider a per-test-method tear down that logs externally:

```python

    def expectThat(self, matchee, matcher, message='', verbose=False):
        """ An assertion helper method.

            Args:
                matchee: An object to match with matcher.
                matcher: An object meeting the testtools.Matcher protocol.
                message (str)
                verbose (bool)

            Returns:
                None
        """
        self.logger.debug('called expectThat, ' + message)
        # broad wrapper around TestTools `expectThat` method for additional
        # conditions and logging.
        super(CouncilTestCase, self).expectThat(matchee, matcher, message, verbose)

    def tearDown(self):
        """ TearDown a test method defined in a TestCase class. """

        self.logger.debug('test-case teardown')

        interfaces = self._result_interfaces()

        if interfaces:
[...]

PR #240, opened 2016-09-17
feature/expectThat-return

blakev:
Feature #239

---

This change is [Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/240)

### DIFF
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -385,10 +385,15 @@ class TestCase(unittest.TestCase):
         Handlers are called before the outcome (such as addFailure) that
         the exception has caused.
 
+        Each handler is passed an `exc_info` tuple that contains the raised
+        Exception, exception parameters and the stack trace.
+
         Handlers are called in first-added, first-called order, and if they
         raise an exception, that will propogate out of the test running
         machinery, halting test processing. As a result, do not call code that
         may unreasonably fail.
+
+        :seealso onException:
         """
         self.__exception_handlers.append(handler)
 
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -529,7 +529,8 @@ class TestCase(unittest.TestCase):
         :param matchee: An object to match with matcher.
         :param matcher: An object meeting the testtools.Matcher protocol.
         :param message: If specified, show this message with any failed match.
-
+        :return: A tuple containing whether the condition succeed or failed,
+            and if it failed the mismatch details explaining why.
         """
         mismatch_error = self._matchHelper(matchee, matcher, message, verbose)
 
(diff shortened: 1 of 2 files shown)


thomir:
Hi,

Thanks for your contribution.

This changes backwards compatibility, which makes it difficult to merge as-is. I've also commented on your bug report with a request for some additional information.

Cheers,

blakev:
Nobody would be using `something = self.expectThat(True, True)` because it would always be `False`. I'd be extremely surprised if this broke anyone else's tests.

testtools/testcase.py
```diff
@@ -536,6 +542,9 @@ def expectThat(self, matchee, matcher, message='', verbose=False):
                 )
             )
             self.force_failure = True
+            return False, str(mismatch_error)
```
rbtcollins: I don't think there is any need for a tuple here: an Option makes more sense, which in idiomatic Python is just None-or-value. So lets just return mismatch_error.

testtools/tests/test_testcase.py
```diff
@@ -1967,6 +1967,39 @@ def test_other_attribute(self):
         self.assertRaises(AttributeError, getattr, orig, 'thing')
 
 
+class TestTCaseExpectThat(TestCase):
```
rbtcollins: Whats TCase?

testtools/tests/test_testcase.py
```diff
@@ -1967,6 +1967,39 @@ def test_other_attribute(self):
...
+
+    def setUp(self):
+        super(TestTCaseExpectThat, self).setUp()
+        self.passed = None
+        self.message = 'default'
+
+    def expectThat(self, matcher, matchee, message='', verbose=False):
```
rbtcollins: I don't think there is any need for this helper to shadow expectThat, and in shadowing it it will surprise folk - as its not a method one would expect to be shadowed.

testtools/tests/test_testcase.py
```diff
@@ -1967,6 +1967,39 @@ def test_other_attribute(self):
...
+        passed, message = super(TestTCaseExpectThat, self).expectThat(matcher, matchee, message, verbose)
+        self.passed = passed
+        self.message = message
+        return passed, message
+
+    def test_expectthat_success(self):
+        s, m = self.expectThat(True, Equals(True))
+        assert s
+        assert s == self.passed
+        assert m is self.message is None
```
rbtcollins: I'd prefer to use richer comparisons than bare asserts: unlike py.test, unittest doesn't currently introspect assert statements.

e.g.

```
    def test_expectThat_success(self):
        self.assertThat(self.expectThat(True, Equals(True)), Is(None))
```

and

```
    def test_expectThat_failure(self):
        mismatch = self.expectThat(False, Equals(True))
        self.assertThat(str(mismatch), Equals(str(Equals(True).match(False))))
```

will give complete coverage I believe.

jelmer:
I'm going to close this since it's gone 5 years without follow-up; feel free to reopen if you're still interested in working on it.

PR #241, opened 2016-09-19
Make KeysEqual usable with no arguments, i.e. match a dict with no keys.

allenap:
This change is [Reviewable](https://reviewable.io/reviews/testing-cabal/testtools/241)

### DIFF
--- a/testtools/matchers/_dict.py
+++ b/testtools/matchers/_dict.py
@@ -4,6 +4,8 @@ __all__ = [
     'KeysEqual',
     ]
 
+from collections import Mapping
+
 from ..helpers import (
     dict_subtract,
     filter_values,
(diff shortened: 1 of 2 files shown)


testtools/matchers/_dict.py
```diff
@@ -235,15 +237,15 @@ class KeysEqual(Matcher):
...
-        :param expected: The keys the dict is expected to have.  If a dict,
-            then we use the keys of that dict, if a collection, we assume it
-            is a collection of expected keys.
+        :param expected: The keys the matchee is expected to have. As a
+            special case, if a single argument is specified, and it is a
+            mapping, then we use its keys as the expected set.
         """
         super(KeysEqual, self).__init__()
-        try:
+        if len(expected) == 1 and isinstance(expected[0], Mapping):
```
freeekanayaka: It seems there's a slight change in behavior too here, the current code was happy with anything implementing a keys() method returning a list. With this branch we'll also require it to be a subclass of Mapping, I think. Perhaps something like:

```python

if len(expected) == and safe_hasattr(expected[0], "keys"):

would be a bit better, from a backward-compatibility point of view.

Comment (allenap) [inline testtools/matchers/_dict.py]: Eek, `Mapping` does not work as I expected it to, so I've removed that, and gone to a simpler approach.

BTW, sorry for the long delay. I must have missed the notification from GitHub and this slipped my mind. Thanks for the review.

File: testtools/matchers/_dict.py [hunk 2]
```diff
@@ -235,15 +237,15 @@ class KeysEqual(Matcher):
...
             self.expected = expected[0].keys()
-        except AttributeError:
-            self.expected = list(expected)
+        else:
+            self.expected = expected
```
allenap: Okay, no probs. I was trying to save a copy, but that can wait for another day.

PR #242, opened 2016-11-26
Make twistedsupport and related tests fully work on Python 3

freeekanayaka:
Now that Python 3 support in Twisted is mature enough, we can have testtools.twistedsupport fully working with Python 3.

This branch fixes some of the magic in _spinner.trap_unhandled_errors (which did not work in with Python 3, as test failures showed), so the suite is now fully green on Python 3:

$ python3 -m testtools.run testtools.tests.test_suite
Tests running...

Ran 2612 tests in 1.311s
OK

$ python -m testtools.run testtools.tests.test_suite
Tests running...

Ran 2612 tests in 1.253s
OK

The hard-coded test dependency on Twisted 13 has been removed as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,21 +7,9 @@ python:
   - "3.5"
   - "pypy"
 
-# Twisted tests currently only work on Python 2.
-matrix:
-  include:
-    - python: "2.7"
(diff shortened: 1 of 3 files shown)


freeekanayaka:
@jml any objection to merging this branch?

PR #248, opened 2017-01-22
Making sure that TestCase can be hashed

gone:
Defining a custom __eq__ method will prevent a TestCase instance from being hashed, and the django testing library attempts to hash all TestCases.

### DIFF
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -276,11 +276,8 @@ class TestCase(unittest.TestCase):
         return self.__dict__ == other.__dict__
 
     def __hash__(self):
-        hashfn = getattr(unittest.TestCase, '__hash__', None)
-        if hashfn is not None:
-            return hashfn(self)
-        return id(self)
-
+        hashfn = getattr(unittest.TestCase, '__hash__', id)
+        return hashfn(self)
 
     def __repr__(self):
         # We add id to the repr because it makes testing testtools easier.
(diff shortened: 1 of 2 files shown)


ddormer:
I don't think the hash should default to `self.id` because that will result in each test having the same hash. Instead `__hash__` should be set to `unittest.TestCase.__hash__`.

javiromero:
I'm facing the exact same issue when running Django 1.11.x tests with Python 3. Is there any chance this gets merged?

ddormer:
Ping @jml and @gone :)

jml:
Unsubscribe

Flimm:
This looks useful to me! While waiting for the merge, I guess I can copy and paste the `__hash__` implementation.

gone:
@ddormer sorry about ghosting - missed the notification for this. 

I see what you mean about using every test within the same case having the same id, but that's only going to be used if  `unittest.TestCase.__hash__` isn't set - so it will prefer unittest.testcase if it's present and only all back to id if it's not there.

gone:
Bumping this up - been more active lately if there's desire to get this merged

testtools/testcase.py
```diff
@@ -276,6 +276,10 @@ def __eq__(self, other):
             return False
         return self.__dict__ == other.__dict__
 
+    def __hash__(self):
+        hashfn = getattr(unittest.TestCase, '__hash__', id)
+        return hashfn(self)
```
cjwatson, line 281: This seems basically reasonable, but it's more complicated than necessary now that testtools no longer supports Python 2.  Could you make this be just:

```python
    __hash__ = unittest.TestCase.__hash__
```

cjwatson (approved):

gone:
Updated - let me know if you'd like me to squash

PR #249, opened 2017-02-07
Parse arguments for TestProgram class

kopecmartin:
The patch allows to pass argument failfast to TestProgram class from main
method. If it's needed, it's possible to add more arguments.
Example of usage: python -m testtools.run --failfast <test>

### DIFF
--- a/testtools/run.py
+++ b/testtools/run.py
@@ -260,8 +260,14 @@ class TestProgram(unittest.TestProgram):
 ################
 
 def main(argv, stdout):
-    program = TestProgram(argv=argv, testRunner=partial(TestToolsTestRunner, stdout=stdout),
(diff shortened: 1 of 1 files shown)


freeekanayaka:
Hi,

I don't understand this branch. The --failfast command option is already parsed and honored by testtools.run.TestProgram (and it's done by re-using the code the stdlib unittest.TestProgram class).

Can you please provide an example where "python -m testtools.run --failfast" currently does not work?

kopecmartin:
Hi, 
I didn't notice that. I retested it and it works as expected. I made a mistake somewhere. Thank you for your time.

freeekanayaka:
Happy it works for you. I'll close this PR then.

PR #252, opened 2017-02-08
Don't add file without content

therve:
In the test record API, skip adding file details to the record if there
are not bytes. This reduces that amount of data kept when there is none.

### DIFF
--- a/testtools/testresult/real.py
+++ b/testtools/testresult/real.py
@@ -832,7 +832,7 @@ class _StreamToTestRecord(StreamResult):
 
         case = case.got_timestamp(timestamp)
 
-        if file_name is not None:
+        if file_name is not None and file_bytes:
             case = case.got_file(file_name, file_bytes, mime_type)
 
         if test_tags is not None:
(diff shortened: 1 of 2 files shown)


freeekanayaka:
Hi Thomas, long time no see, hope you're good!

How hard would it be to add a unit test for the change? I didn't yet look at existing unit-test coverage for that part of the code.

Also, a couple of questions about the issue:

- Is that 25M the size of the subunit artifact for a single run or for N runs?
- Did you already verify that the fix in this branch shaves that size significantly? How much?

Cheers

therve:
Hi Free! Good to hear from you too.

I haven't looked at tests yet, I'll see how's that looking.

- The 25M is for a single run. See for example http://logs.openstack.org/56/430056/1/check/gate-heat-python27-db-ubuntu-xenial/41e6c88/testrepository.subunit.gz for the compressed subunit version (50% uncompressed).
- The fix reduces the size to about 4M.

freeekanayaka:
I've modified [subunitdetails](https://pypi.python.org/pypi/subunitdetails/0.0.4) slightly to inspect the subunit artifact linked above by Thomas.

There 278024 empty file attachments (I didn't check the type, but presumably it's mostly logging indeed). If we consider an average subunit record size of 75 bytes per attachment (mostly the content-type string plus test ID string), that adds up to about 20M which confirms the analysis.

PR #268, opened 2018-04-18
Allow skip/skipIf/skipUnless as class decorators

### LINKED ISSUE #205: testtools.skipIf doesn't work as a class decorator
I tried decorating a `testtools.TestCase` with `testtools.skipIf` and the tests in that class still ran.

I'll try and provide a simple test to reproduce the problem.

zaneb:
In unittest/unittest2 the skip/skipIf/skipUnless decorators can be used to
decorate either individual test methods or entire TestCase classes.
However, the testtools equivalents could previously only be used to
decorate test methods. If used on a TestCase the class would be replaced
with a function, resulting in the tests either not be discovered at all or
(if a custom test loader was used) potentially an error.

This change allows the skip decorators to be used on the TestCase subclass,
for equivalent functionality with unittest.

Fixes #205

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,16 @@ testtools NEWS
 
 Changes and improvements to testtools_, grouped by release.
 
+NEXT
+~~~~
+
+Improvements
+------------
+
+* The skip, skipIf, and skipUnless decorators can now be used as class
+  decorators as well as test method decorators, just as they can in
+  unittest.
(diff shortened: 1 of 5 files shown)


zaneb:
The build failed on Python 3.3 because Sphinx is no longer available for Python 3.3. Other than that, everything passed.

PR #269, opened 2018-04-18
Don't install Sphinx on Travis Python 3.3 target

zaneb:
Sphinx doesn't support Python 3.3, and we already don't run Sphinx in the
Travis Python 3.3 target. However, it fails to install because there is no
longer any Sphinx package available for Python 3.3. So avoid even trying to
install it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,8 @@ python:
 
 install:
   - pip install -U pip wheel setuptools
-  - pip install sphinx Twisted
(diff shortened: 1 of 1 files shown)


zaneb:
Yeah, I just noticed there's another PR (#266) to drop support for 3.3 altogether.

rbtcollins:
Merged the 3.3 removal patch

Issue #270, opened 2018-05-12
Tests fail on Python 3.7


hroncok:
Hi, I'm preparing an update of Python 3 on Fedora to Python 3.7. Currently, testtools fails to build with Python 3,7, because of a test failure. I've managed to reproduce it on Travis:

https://travis-ci.org/hroncok/testtools/jobs/378123060

```
$ python -m testtools.run testtools.tests.test_suite
/opt/python/3.7-dev/lib/python3.7/runpy.py:125: RuntimeWarning: 'testtools.run' found in sys.modules after import of package 'testtools', but prior to execution of 'testtools.run'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
Tests running...
======================================================================
FAIL: testtools.tests.matchers.test_exception.TestMatchesExceptionInstanceInterface.test__str__
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/hroncok/testtools/testtools/tests/matchers/helpers.py", line 26, in test__str__
    self.assertThat(matcher, DocTestMatches(expected))
  File "/home/travis/build/hroncok/testtools/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: Expected:
    MatchesException(Exception('foo',))
Got:
    MatchesException(Exception('foo'))
======================================================================
FAIL: testtools.tests.matchers.test_exception.TestMatchesExceptionInstanceInterface.test_describe_difference
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/hroncok/testtools/testtools/tests/matchers/helpers.py", line 33, in test_describe_difference
    self.assertEqual(difference, mismatch.describe())
  File "/home/travis/build/hroncok/testtools/testtools/testcase.py", line 411, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/home/travis/build/hroncok/testtools/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: !=:
reference = "ValueError('bar',) has different arguments to ValueError('foo',)."
actual    = "ValueError('bar') has different arguments to ValueError('foo')."
Ran 2625 tests in 1.219s
FAILED (failures=2)
```

Travis CI runs 3.7.0a4+, I have the same results with 3.7.0b4.

hroncok:
It's a change in repr:

```pycon
Python 3.6.5 ...
[...]

PR #271, opened 2018-05-17
Fix the tests on Python 3.7

hroncok:
Exception's repr got changed not to include trailing comma

Fixes https://github.com/testing-cabal/testtools/issues/270


hroncok:
Is there anything I can do to move this forward?

Issue #272, opened 2018-06-29
unittestt.SkipTest is treated as failure in setUpClass() by testtools.run


mtreinish:
When running a test suite built with stdlib unittest (running with Python 3.6.5) if a setUpClass() method raises a unittest.SkipTest exception to skip the test the testtools.run runner is treating that as an error. For example using a synthetic example like:

```
import unittest

class TestDirectSkip(unittest.TestCase):

    @classmethod
    def setUpClass(cls):
        super(TestDirectSkip, cls).setUpClass()
        cls.test = 'a'
        raise unittest.SkipTest("I really skip")

    def test_foo(self):
        self.assertEqual(self.test, 'a')
```
will return a failure like:

```
Tests running...
======================================================================
ERROR: setUpClass (test_skips.TestDirectSkip)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/tests/test_skips.py", line 29, in setUpClass
    raise unittest.case.SkipTest("I really skip")
unittest.case.SkipTest: I really skip
```

My thinking it it has something to do with: https://github.com/testing-cabal/testtools/blob/master/testtools/testcase.py#L66-L69 because if you switch the exception from unittest to unittest2 in the example like:

```
import unittest

class TestDirectSkip(unittest.TestCase):

    @classmethod
    def setUpClass(cls):
        super(TestDirectSkip, cls).setUpClass()
        cls.test = 'a'
        import unittest2
        raise unittest2.SkipTest("I really skip")

    def test_foo(self):
        self.assertEqual(self.test, 'a')
```
The skip is properly recognized.

rbtcollins:
Interesting. Can you check the actual exception types at play here?

mtreinish:
The reverse case is also an issue here. The `unittest2.case.SkipTest` raised by testtools for `skipException` is treated as a failure by stdlib unittest. For example, running:

```
import testtools

class TestDirectSkip(testtools.TestCase):

    @classmethod
    def setUpClass(cls):
        super(TestDirectSkip, cls).setUpClass()
        cls.test = 'a'
        raise self.skipException("I really skip")

    def test_foo(self):
        self.assertEqual(self.test, 'a')
```

with `python -m testtools.run` will skip as expected. But when you run it with `python -m unittest` you
get a failure like:

```
E
======================================================================
ERROR: setUpClass (test_skip.TestDirectSkip)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #277, opened 2019-05-03
Remove unittest2 usage from testtools

### LINKED ISSUE #263: Remove use of obsolete unittest2
From the unittest2 PyPI page:

> unittest2 is a backport of the new features added to the unittest testing framework in Python 2.7 and onwards.

testtools no longer supports Python <2.7 so there is no reason to use the backport unittest2 module.  Instead, the standard library unittest module can be used.

This would be great since unittest2 is old, crufty, and essentially unmaintained.

mtreinish:
Currently testtools bases all of it's unittest extensions off of
unittest2 instead of the stdlib unittest. At one point this made sense
since unittest2 provided a stable base as unittest in stdlib added
features. But it's been ~5 years since there was a unittest2 release
(or a patch merged) and things have changed since then. The best
example of this is of the supported python versions listed in the
unittest2 project description/README only one is still supported by
upstream python, 2.7, which goes end of life at the end of this year.

More specific to testtools the use of unittest2 causes a whole slew of
issues because of differences in behavior with stdlib unittest. For
example here a couple issues encountered:

https://bugs.launchpad.net/testtools/+bug/1467558
https://bugs.launchpad.net/testtools/+bug/1417803
mtreinish/stestr#238
testing-cabal/testtools#272

which are caused, at least in part, by unittest2. There are likely other
bugs related to it that haven't been reported (or I just missed/forgot
about). At this point it's better to remove the unittest2 usage and just
rely on the upstream stdlib unittest which if nothing else is actively
maintained. It'll improve compatibility using the testtools runner with
stdlib unittest test suites and removes the class of bugs caused by
the differences in unittest2.

Fixes #263


testtools/testcase.py
```diff
@@ -497,6 +492,23 @@ def assertThat(self, matchee, matcher, message='', verbose=False):
         if mismatch_error is not None:
             raise mismatch_error
 
+    def assertItemsEqual(self, a, b, message=''):
```
hugovk: Is this needed in this form?

`assertItemsEqual` is available in Python 2.7 and is named `assertCountEqual` in Python 3.

https://docs.python.org/2/library/unittest.html

Can they be used directly? 

eg. 

```python
try:
    # Python 2.7
    from unittest import assertItemsEqual as assertCountEqual
except ImportError:
    # Python 3
    from unittest import assertCountEqual
```

hugovk:
+1

`unittest2` is needed for Python 2.6, but that's no longer supported so it's better to use the stdlib `unittest `.

stmcginnis:
Would be great to get this in. I ran into this over the last couple days, and took me quite awhile to track it down to the same conclusion and find this issue and PR.

mtreinish:
> +1
> 
> `unittest2` is needed for Python 2.6, but that's no longer supported so it's better to use the stdlib `unittest `.

It's also needed for python 2.7 support, unittest2 was a rolling backport and included some of the test runner (mainly the loader) from stdlib added after 2.7. (I hit a similar issue in: https://github.com/mtreinish/stestr/pull/256 ). This is why the py2.7 CI jobs are failing here. I've opened up https://github.com/testing-cabal/testtools/issues/286 to discuss dropping 2.7 support. Once we stop supporting 2.7 in testtools this will be unblocked.

mtreinish:
I'm not sure why travis has stopped reporting here, but now that we've dropped support for python 3.4 and python 2.7 this is passing: https://travis-ci.org/github/testing-cabal/testtools/builds/671013953

hugovk:
Maybe disabling and re-enabling the Travis integration will fix it: https://twitter.com/pganssle/status/1246437422726107137

graingert:
perhaps it would be worth switching from travis to github actions?

hugovk:
Please see PR https://github.com/testing-cabal/testtools/pull/290 to add testing on GHA.

testtools/testcase.py
```diff
@@ -497,6 +492,23 @@ def assertThat(self, matchee, matcher, message='', verbose=False):
...
+        Asserts that each element has the same count in both sequences.
+        Example:
+            - [0, 1, 1] and [1, 0, 1] compare equal.
+            - [0, 0, 1] and [0, 1] compare unequal.
+        """
+        if hasattr(self, 'assertCountEqual'):
+            self.assertCountEqual(a, b, message)
+        else:
+            self.assertEqual(collections.Counter(iter(a)),
+                             collections.Counter(iter(b)))
```
graingert, line 506: return six.assertCountEqual(self, a, b, message)

testtools/testcase.py
```diff
@@ -497,6 +492,23 @@ def assertThat(self, matchee, matcher, message='', verbose=False):
...
+
```
graingert, line 507:
```suggestion
    if not six.PY2:
        assertItemsEqual = unittest.TestCase.assertCountEqual

```

graingert: this makes assertCountEqual and assertItemsEqual work on py2 and py3 - I assume that's what you wanted?

stmcginnis: I think it would be good to try to avoid introducing six usage if possible to make it easier when py2 support is dropped. Which getting rid of unittest2 is a step towards.

graingert: @stmcginnis updated to use sys.version_info

imho it's best to use six - then we can remove it with pyupgrade

stmcginnis: True. I'm sure there will be other cleanup work to do.

mtreinish: This actually was leftover from a rebase. Testtools master doesn't support python2 anymore the min python version is 3.5, so I need to remove the if else here

graingert: so you think
```suggestion
    assertItemsEqual = unittest.TestCase.assertCountEqual
```

graingert: @mtreinish ^

testtools/testcase.py
```diff
@@ -16,6 +16,7 @@
     'unique_text_generator',
     ]
 
+import collections
```
FedericoRessi, line 19: Why is this imported here?

testtools/run.py
```diff
@@ -15,7 +15,7 @@
 
 from extras import safe_hasattr, try_imports
 # To let setup.py work, make this a conditional import.
```
FedericoRessi, line 17: I think this comment should go away.

testtools/run.py
```diff
@@ -15,7 +15,7 @@
 
 from extras import safe_hasattr, try_imports
```
FedericoRessi, line 16: It looks like after this change try_imports is no more used

FedericoRessi:
I just found a reference to unittest2 in setup.cfg file that should be removed too.

FedericoRessi:
Can you please have a look at differences[1] between this change and PR #295? I think you are still missing something here while #295  is up to date with master branch and it is passing CI verifications.

Cheers

[1] https://github.com/mtreinish/testtools/compare/remove-unittest2...FedericoRessi:drop-unittest2

ssbarnea (approved):
Highly needed as unittest2 should have never being installed for newer python versions. It is causing a very long list of warnings with newer pythons.

ssbarnea:
@mtreinish time to press the update branch.

asears:
A warning coming up that would likely be removed after this PR.

```python
c:\tools\anaconda3\lib\site-packages\unittest2\compatibility.py:143
  Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

ssbarnea:
@mtreinish Any chance to add me to the project in order to avoid these to stall even longer?

dtantsur:
@mtreinish could you make a release with this change? The latest release is broken on Python 3.10 because of unittest2.

mtreinish:
@dtantsur I would, but I unfortunately do not have access to the project on pypi otherwise I would have released already. I'll try to reach out to @rbtcollins and/or @jelmer and fix that so we can get the release out ahead of the 3.10 release

dtantsur:
Great to know, thanks! FYI Fedora Rawhide has already switched to 3.10, so everything using testtools may fail to build.

ssbarnea:
@mtreinish If you do not get an answer quickly you can raise a ticket on https://github.com/pypa/pypi-support/issues/new/choose asking to be added as owner. You can also mention me in the ticket and I will back your request. Usually pypa folks are quite fast on helping with tickets like this.

rbtcollins:
@ssbarnea they are responsive, but they also don't take over non-orphaned projects. testtools has 5 people capable of doing releases - me; @freeekanayaka @jelmer @jml @thomir :- perhaps some are no longer interested in the project in which case they can remove themselves.

@mtreinish you're mtreinish on pypi ?

mtreinish:
@rbtcollins yeah, that's my pypi username. (https://pypi.org/user/mtreinish/ )

rbtcollins:
@mtreinish done

PR #278, opened 2019-06-19
PyCharm IDE unittest detection compatibility fix

rm-you:
PyCharm is a Python IDE developed by JetBrains. It uses some algorithm
to detect classes that are derived from unittest.TestCase, and the
dynamic import that is used prior to this patch causes it to
malfunction. It fails to detect tests that are derived from
testtools.TestCase as unittests for the purposes of their runner.
Doing the conditional imports in a more standard Pythonic way, we can
allow testtools to be more compatible with their approach.

I believe that it should not necessarily be the responsibility of every
library author/maintainer to go out of their way to adjust to the
(flawed/limited) functions of various IDEs/toolchains, but in this case
I think the change is minimal and Pythonic, and the benefit is
significant for those who use this fairly popular IDE.

For the original bug on their tracker, see:
https://youtrack.jetbrains.com/issue/PY-26630

### DIFF
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -25,10 +25,14 @@ import warnings
 from extras import (
     safe_hasattr,
     try_import,
-    try_imports,
     )
 # To let setup.py work, make this a conditional import.
-unittest = try_imports(['unittest2', 'unittest'])
+# Don't use extras.try_imports, as it interferes with PyCharm's unittest
+# detection algorithm. See: https://youtrack.jetbrains.com/issue/PY-26630
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
 import six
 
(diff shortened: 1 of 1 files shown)


rm-you:
This would also be resolved by https://github.com/testing-cabal/testtools/pull/277 but I don't know if that will be mergeable any time soon...

jelmer:
Seems reasonable to me.

rm-you:
Talked to mtreinish, and it seems like #277 will be a while yet, so merging this in the meantime would be helpful. :)

jelmer:
I'll wait a couple of days to allow others the opportunity to object; if there are no objections I'll merge some time next week.

rm-you:
Thanks. :)

PR #280, opened 2019-09-18
Publicly expose the SameMembers matcher

bigjools:
This is a useful matcher that is not currently exposed in the public
matchers module from its 'private' _basic.py module. I'm sure this was
just an oversight by the original author, so this change exposes it
properly in matchers.py.


mtreinish (approved):

bigjools:
Thanks Matthew \o/

PR #281, opened 2019-11-10
Resolve DeprecationWarnings under python 36

pvinci:


testtools/content.py
```diff
@@ -102,7 +102,7 @@ def iter_text(self):
         no charset parameter is present in the MIME type. (This is somewhat
         arbitrary, but consistent with RFC2617 3.7.1).
 
-        :raises ValueError: If the content type is not text/\\*.
+        :raises ValueError: If the content type is not text *.
```
jelmer: Perhaps "text/*" ? Replacing it with a space makes it unparseable to me.

ssbarnea:
I think that any core can fix this PR. Anyone?

testtools/content.py
mtreinish, line 105:
```suggestion
        :raises ValueError: If the content type is not "text/*".
```

ssbarnea (approved):
LGTM

mtreinish (approved):

PR #288, opened 2020-01-18
Update docs about which version to use for Python 3.4

hugovk:
https://github.com/testing-cabal/testtools/pull/284 dropped support for EOL Python 3.4.

This PR documents it, in only one place to remove duplication, and refer to that one place from the other.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,11 +33,12 @@ Supported platforms
 
  * Python 2.7 or 3.5+ / PyPy
 
-If you would like to use testtools for earlier Pythons, please use testtools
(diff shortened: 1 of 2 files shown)


stmcginnis:
Looks great to me. I think it would be good to get the documentation updated for this since the current documentation is not clear on the latest changes.

stmcginnis:
Good point. It does clarify immediately after which versions would be needed, but since it is not inclusive of the minimum version mentioned by the caps, that line should probably just state the current support with the details following being instructions what to do if you need something older.

hugovk:
Updated!

PR #290, opened 2020-04-14
Test on GitHub Actions

hugovk:
Travis CI is playing up (https://github.com/testing-cabal/testtools/pull/277#issuecomment-609407671), so as suggested in https://github.com/testing-cabal/testtools/pull/277#issuecomment-613297206 let's test using GitHub Actions.

This does the same as `.travis.yml`

Here's an example run: https://github.com/hugovk/testtools/actions/runs/77822864


hugovk:
OK, deleted!

mtreinish:
There aren't any checks run here? If we're going to switch to gh actions for CI then we should actually be running it before we delete the travis config. Travis reported on the first commit on this branch but since deleting the travis config there are no checks shown.

graingert:
> There aren't any checks run here? If we're going to switch to gh actions for CI then we should actually be running it before we delete the travis config. Travis reported on the first commit on this branch but since deleting the travis config there are no checks shown.

I think GHA needs enabling by an admin

hugovk:
Or I think it should be active once the GHA config file is merged.

Shall I revert the `.travis.yml` removal from this PR, so it can be merged and verified first?

mtreinish:
I've added back the travis config lets merge this and get it running (from what I can tell actions are enabled on the repo already) then once it's confirmed working we can remove the travis config.

Issue #235, opened 2016-06-16
Doesn't support subTests


elopio:
python's unittest now has a cool feature for test scenarios called subTest: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

I'm trying to update some tests from unittest.TestCase to testtools.TestCase, but there are some that use subTest which blocks me from using testtools.

This is the trace from one of the failing tests:
======================================================================
ERROR: test_yaml_valid_epochs (snapcraft.tests.test_yaml.TestYaml)
snapcraft.tests.test_yaml.TestYaml.test_yaml_valid_epochs
----------------------------------------------------------------------
testtools.testresult.real._StringException: Empty attachments:
  pythonlogging:''
  pythonlogging:''-1

Traceback (most recent call last):
  File "/usr/lib/python3.5/unittest/mock.py", line 1157, in patched
    return func(*args, **keywargs)
  File "/home/elopio/workspace/canonical/snapcraft/snapcraft/tests/test_yaml.py", line 669, in test_yaml_valid_epochs
    with self.subTest(key=epoch):
  File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
    return next(self.gen)
  File "/usr/lib/python3/dist-packages/unittest2/case.py", line 535, in subTest
    if not self._outcome.result_supports_subtests:
AttributeError: 'NoneType' object has no attribute 'result_supports_subtests'

I think it would be great if testtools supports subTests.

rbtcollins:
Indeed, it would be nice, Would love it if you wanted to put a patch forward.

rbtcollins:
Arguably the bug is in subTest itself - it presumes _outcome exists, but _outcome is a unittest internal attribute that isn't part of the API contract, so we should fix this in CPython's stdlib.

elopio:
I agree that unittest should have a fall back, and that the error sucks because of what they are doing. I'll rename this bug.

elopio:
Oh, noes, it was you who renamed it. I think my name was better :smile:

nicoddemus:
https://github.com/python/cpython/pull/9707 👍

nicoddemus:
That PR has been merged, so I guess this can now be closed. 👍

marph91:
I'm running in the same error, using Python 3.6.9. The change should be already included in 3.6.8.

Simple reproducer:

`testsuite.py`
```python
import testtools


class TC(testtools.TestCase):
    def test_subtest(self):
[...]

PR #295, opened 2020-07-08
Drop unittest2 in favor of newer unittest implementations

FedericoRessi:
unittest2 library isn't been updated since some year.

Newer python frameworks are expected to has far more fixes than that unittest2.

Actually unittest2 documentation states it is only tested up to python 3.4

This therefore limit the use of unittest2 library up to the last python 3.4 release.

It is expected to fix Issue #235 when recent enough version of Python is installed


stmcginnis:
I'm all for this change, but I think it will take a little more work.

There are places in the code that expect unittest2 to be present:

https://github.com/testing-cabal/testtools/blob/83cb1545e0b409e3cf7fbab6eff7a333713822cd/testtools/testsuite.py#L38

With this change, unittest2 will only be installed in old python versions. Newer versions will have an issue at runtime when it tries to load classes like these.

We also limit installation to python 3.5 or great:

https://github.com/testing-cabal/testtools/blob/83cb1545e0b409e3cf7fbab6eff7a333713822cd/setup.py#L14

So this makes it kind of a no-op.

I think ideally we should phase out the use of unitttest2. There are other issues like https://github.com/testing-cabal/testtools/issues/272#issuecomment-573860489 where the consuming code needs to handle the appropriate unittest or unittest2 code. So if we change this, we need to make sure it is released with a major version bump so it isn't entirely unexpected in consuming projects if things start breaking for them.

mtreinish:
Yeah, I agree with @stmcginnis here, we've moved to only supporting python 3.5, I think the best path forward here is to just use unittest everywhere and remove all the unittest2 usage. This is in progress (albeit slowly because I keep getting sidetracked with other work) in #277

stmcginnis:
We should probably stick with https://github.com/testing-cabal/testtools/pull/277

FedericoRessi:
@mtreinish I merged #277 into this to get what I was missing. Please merge this into your #277 so you get what you missed.
If you look well in run.py you still have a reference to unittest2 around line 47. Cheers

tox.ini
```diff
@@ -0,0 +1,18 @@
+[tox]
+envlist = py35,py36,pypy,pypy3
```
hugovk: Python 2 is no longer supported, and 3.7 and 3.8 are supported.

```suggestion
envlist = py35,py36,py37,py38,pypy3
```

mtreinish:
I've pulled this into #277 so I'm going to close this

PR #301, opened 2020-12-02
Fix github actions

ssbarnea:


hugovk:
Looks like the problem is a conflict in dependencies, triggered by the new pip dependency resolver:

https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020

For a quick and temporary fix, use the `--use-deprecated=legacy-resolver` flag, but that's only available until the next month or so.

mtreinish (commented):
I think the issue here is that the checkout doesn't have the full git history so the pbr git integration doesn't have tags to use for the version string. Can you update the checkout task to be:
```yaml
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0
```
to do a full clone and see if that fixes it

ssbarnea:
> I think the issue here is that the checkout doesn't have the full git history so the pbr git integration doesn't have tags to use for the version string. Can you update the checkout task to be:
> 
> ```yaml
>       - uses: actions/checkout@v2
>         with:
>           fetch-depth: 0
> ```
> 
> to do a full clone and see if that fixes it

Yep. Realised that but got into long zoom call. I had the same problem with setuptools-scm, which can fail to generate a version when not having git history.

mtreinish (approved):

PR #302, opened 2021-02-11
Reduce dependencies

stephenfin:
I want to get to a point where fixtures and testtools don't form a circular
dependency chain. Before we can get there though, there's a lot more things
that appear to need cleaning up, namely our dependency on testtools2, extras,
and mimeparse, all of which are effectively EOL.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -25,6 +25,11 @@ Changes
   moved to the ``extras`` package. It is no longer useless on Python 3-only
   projects.
 
+* The ``try_imports`` utility has been removed from ``testtools.helpers``.
+  This was a compat wrapper introduced in 0.9.25 when the utility itself was
+  moved to the ``extras`` package. It is no longer used within testtools and
+  has therefore been dropped.
+
 2.4.0
 ~~~~~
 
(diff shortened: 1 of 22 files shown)


stephenfin:
This is passing locally, so I've no idea what's up here :disappointed:

mtreinish:
@stephenfin I just merged #277 which removed the unittest2 dependency could you rebase this to do the other requirement cleanup (extras and mimeparse)? Once it's rebased if it's still failing i'll take a look and see if I can reproduce the failures locally.

stephenfin:
> @stephenfin I just merged #277 which removed the unittest2 dependency could you rebase this to do the other requirement cleanup (extras and mimeparse)? Once it's rebased if it's still failing i'll take a look and see if I can reproduce the failures locally.

Done. Turns out the failure was because I was removing `extras` but `fixtures` has an indirect dependency on this via `testtools`. I've pushed https://github.com/testing-cabal/fixtures/pull/46 to drop support for that. If you could take a look at this also, I'd be very appreciative. Eventually I'd like to uncouple `testtools` and `fixtures`, such that the former depends on the latter but not the other way around, but there's a bit of work to go for that yet.

stephenfin:
Hurrah. Can we get a new release of this and fixtures so I can proceed with dropping additional dependencies?

PR #311, opened 2021-06-16
testtools.compat: provide BytesIO, StringIO again

mhuin:
This allows projects dependent on testtools.compat some respite if
they haven't been updated for python 3.10 yet.

### DIFF
--- a/testtools/compat.py
+++ b/testtools/compat.py
@@ -8,6 +8,8 @@ __all__ = [
     'advance_iterator',
     'reraise',
     'unicode_output_stream',
+    'StringIO',
+    'BytesIO',
     ]
 
 import codecs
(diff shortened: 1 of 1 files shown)


testtools/compat.py
```diff
@@ -16,6 +18,8 @@
 import os
 import sys
 import unicodedata
+# ensure retrocompatibility < 3.10
```
mtreinish, line 21:
```suggestion
# Ensure retro-compatibility with older testtools releases
```

mtreinish (approved):
This seems fair enough to me, these were there to have a consistent api with py3 vs py2 but we haven't supported py2 for a while now. But I have seen issues around this outside so better to avoid the breaking change, just a small tweak to the comment text and this is good to go

PR #319, opened 2022-01-10
Upgrade asserts with teyit for Python 3.11 compatibility

hugovk:
Use https://github.com/isidentical/teyit to upgrade unittest asserts for better error messages.

Also `assertEquals` has been removed in 3.11, so use `assertEqual` instead.

https://docs.python.org/3.11/whatsnew/3.11.html

---

There are still some other things needed for 3.11 compatibility but this is a first step.


hroncok:
What does "teyit" mean?

hugovk:
It's this handy tool: https://github.com/isidentical/teyit

stephenfin (approved):

hugovk:
Hmm, strange the flag looks enabled for me:

image

Anyway, rebased!

cjwatson (approved):

PR #338, opened 2022-12-27
Fix distutilscmd in Debian + Py 3.11

thomasgoirand:
Under Debian, when attempting to run tests, the debian folder is found and conflicts with this error:

setuptools.errors.PackageDiscoveryError: Multiple top-level packages discovered in a flat-layout: ['debian', 'testtools'].

This patch fixes this (I already applied it in Debian, and it solved the issue for me).


jelmer (approved):

jelmer:
Thanks!

PR #351, opened 2023-09-17
Migrate to pyproject.toml

jelmer:
Migrate to pyproject.toml


jelmer:
Alas, it looks like pbr doesn't support project config in pyproject.toml yet :-/

bigjools:
FWIW I switched a few of my projects to `pyproject.toml` and removed `pbr` in the process. I changed my build system to `hatch` and added this config

```
[tool.hatch.version]
source = "vcs"

[tool.hatch.build.hooks.vcs]
version-file = "_version.py"
```

You still get the auto-versioning based on Git tag. You can also use 
`[tool.hatch.build.targets.wheel.shared-data]`
to add extra data files in the wheel.

I think `pbr` has run its course and should go now.

jelmer:
Hmm, hatch looks interesting - and I think I've come across it in other projects.

Doesn't appear to be packaged in Debian yet, so perhaps I'll have a look at that first.

bigjools:
Here's one of mine if you want to see what we're doing with it. https://github.com/juledwar/soufi/
I'm not using the test matrix it can do (still using GH actions for that) but it's SO useful being able to do things like `hatch run tests` or `hatch run python3`

PR #353, opened 2023-09-22
Reduce use of deprecated `distutils` module

### LINKED ISSUE #352: `testtools` fails to import with Python 3.12
At the moment, `import testtools` fails because of the eager import of `distutils` via `testtools.distutilscmd`, which provides the deprecated `TestCommand`, since `distutils` is removed from the Python 3.12 standard library.

`setuptools` still packages `distutils`, so a temporary workaround could be to add `setuptools` as a runtime dependency until `TestCommand` can be removed, or potentially the import of `TestCommand` in the root could be conditional on Python version?

Happy to provide a patch in either direction, if it would help.

jakelishman:
Per PEP 632, `distutils` is deprecated and most uses of it can be replaced by `setuptools`.  The only use within `testtools` is to support the deprecated `TestCommand` that's due to be removed in the next major, but in the meantime, adding a runtime dependency on `setuptools` for Python 3.12+ (which removed `distutils` entirely) allows it to continue working.

Fix #352.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,6 +25,11 @@ classifier =
     Topic :: Software Development :: Libraries :: Python Modules
     Topic :: Software Development :: Testing
 
+[options]
+install_requires =
+    # Purely to support the deprecated `TestCommand` until 3.0.
+    setuptools; python_version>='3.12'
+
 [extras]
 test =
   testscenarios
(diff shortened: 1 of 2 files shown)


philbucher:
Hi, thanks for the fix!
Do you already have plans for a release with this fix?
Thanks!

PR #354, opened 2023-09-22
Migrate from pbr to hatch

jelmer:
Migrate from pbr to hatch


setup.py
```diff
@@ -11,7 +11,8 @@
...
 setuptools.setup(
-    python_requires='>=3.6',
-    cmdclass=cmd_class,
-    setup_requires=['pbr'],
-    pbr=True)
+    packages=setuptools.find_packages(
+        where='.',
+        include=['testtools*'],
+    ),
+    cmdclass=cmd_class)
```
bigjools, line 18: Is there any reason to keep `setup.py` now? You already have a:
```
[tool.setuptools.packages.find]
include = ["testtools"]
```
section and you should be able to also do this in `pyproject.yaml`:

```
[tool.hatch.build]
include = ["/testtools"]

[tool.hatch.build.targets.sdist]
include = ["/testtools"]
```

(I've not tried this so you might need to tweak paths etc)

requirements.txt
```diff
@@ -1,2 +1,3 @@
-pbr>=0.11
+hatchling
+hatch_vcs
```
bigjools, line 2: You can put all these in `pyproject.yaml`.

tox.ini
```diff
@@ -1,11 +1,13 @@
...
+isolated_build = True
 
 [testenv]
 usedevelop = True
 deps =
   sphinx
+  setuptools-scm
 extras =
   test
   twisted
```
bigjools, line 13: It's probably a bit much for this change but I recommend getting rid of `tox` and using Hatch's testing and development venv management.

setup.py
jelmer: We still need to register TestCommand. It's probably possible to do that via pyproject, but I'd rather just punt on that since we've said we'll drop TestCommand anyway at the next major release.

requirements.txt
cclauss: toml, not yaml ;-)

Can this be rebased now that the tests have been updated?

tox.ini
jelmer: That makes sense, but I indeed think we should punt on it.

requirements.txt
jelmer: It's been rebased.

PR #356, opened 2023-10-16
Fix pypy CI runs

jelmer:
Fix pypy CI runs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy]
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3]
+        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy]
```
cclauss, line 11:
```suggestion
        python-version: [3.6, 3.7, 3.8, 3.9, "3.10", 3.11, 3.12, pypy3.10]
```

cclauss:
This changed the name of the `Required` GitHub Action from `build (pypy)` to `build (pypy-3.10)` which means that GHA runs will not complete until a rename is made in the left panel of https://github.com/testing-cabal/testtools/actions
* https://docs.github.com/en/actions/using-workflows/required-workflows#viewing-workflow-runs-for-required-workflows

PR #370, opened 2023-11-24
Format code with ruff

jelmer:


bigjools:
It's a bit tragic that `ruff` doesn't have `black`'s setting `skip-string-normalization`.

jelmer:
I'm actually quite a fan of string normalization, but could be possibly be persuaded to port skip-string-normalization to ruff.

bigjools:
Yeah I don't mind it too — except when you completely destroy the current `git blame` on a mature code base.

jelmer:
I'll merge sometime in 2024 if there are no objections.

stephenfin:
> Yeah I don't mind it too — except when you completely destroy the current `git blame` on a mature code base.

Just create a `.git-blame-ignore-revs` file. GitHub respects that by default and you can configure your local Git to do the same. I stick this header into the file each time I create one:

```
# You can configure git to automatically use this file with the following config:
# git config --global blame.ignoreRevsFile .git-blame-ignore-revs
```

bigjools:
> Just create a `.git-blame-ignore-revs` file. [GitHub respects that](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) and you can configure your local Git to do the same. I stick this header into the file each time I create one:
> ```
> # You can configure git to automatically use this file with the following config:
> # git config --global blame.ignoreRevsFile .git-blame-ignore-revs
> ```

TIL!

bigjools (approved):

Issue #372, opened 2024-05-07
Pytest 8: AttributeError: 'TestExpectedException' object has no attribute 'runTest'. Did you mean: 'subTest'?


hrnciar:
Hello,

we are using pytest for running bandit and python-etcd3gw tests. Since pytest 8 the tests are failing with `AttributeError: 'XXX' object has no attribute 'runTest'. Did you mean: 'subTest'?` 
See:

- https://download.copr.fedorainfracloud.org/results/thrnciar/pytest/fedora-rawhide-x86_64/07388287-bandit/builder-live.log.gz

- https://download.copr.fedorainfracloud.org/results/thrnciar/pytest/fedora-rawhide-x86_64/07424137-python-etcd3gw/builder-live.log.gz

Reproducer:
~~~~
$ git clone git@github.com:testing-cabal/testtools.git
$ cd testtools
[testtools (master)]$ tox -e py312
...

[testtools (master)]$ .tox/py312/bin/pip install 'pytest>=8'
[testtools (master)]$ .tox/py312/bin/pytest testtools/tests
...
AttributeError: 'TestExpectedException' object has no attribute 'runTest'. Did you mean: 'subTest'?
11 warnings, 218 errors

[testtools (master)]$ .tox/py312/bin/pip install 'pytest<8'
[testtools (master)]$ .tox/py312/bin/pytest testtools/tests
...
38 failed, 1328 passed, 7 skipped, 2 xfailed, 59 warnings
~~~~

jelmer:
Please use either the test tools test runner or the native python test runner (unittest)

We're open to PRs that improve support for running the test with third party runners like pytest, but it's not a priority for the project.

jelmer:
See also #332

ncopa:
Details why this happens is found here https://github.com/pytest-dev/pytest/issues/12263#issuecomment-2081434468

PR #373, opened 2024-05-13
Treat methodName="runTest" similar to unittest.TestCase

ncopa:
pytest 82. relies on a feature of `unittest.TestCase` where the initialization of it with the default `methodName="runTest"` is treated specially, allowing it to insantiate even without `runTest` method actually existing.

See under "Changed in Python 3.2" in unittest.TestCase docs

This fixes the error with pytest 8.2:
  AttributeError: 'TestUtil' object has no attribute 'runTest'. Did you mean: 'subTest'?

Fixes: https://github.com/testing-cabal/testtools/issues/372
ref: https://docs.python.org/3/library/unittest.html#unittest.TestCase
ref: https://github.com/python/cpython/blob/51aefc5bf907ddffaaf083ded0de773adcdf08c8/Lib/unittest/case.py#L419-L426
ref: https://github.com/pytest-dev/pytest/issues/12263#issuecomment-2081434468

### DIFF
--- a/testtools/testcase.py
+++ b/testtools/testcase.py
@@ -735,7 +735,17 @@ class TestCase(unittest.TestCase):
 
     def _get_test_method(self):
         method_name = getattr(self, "_testMethodName")
-        return getattr(self, method_name)
+        try:
+            m = getattr(self, method_name)
+        except AttributeError:
+            if method_name != "runTest":
+                # We allow instantiation with no explicit method name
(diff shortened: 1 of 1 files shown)


jelmer (approved):

jelmer (changes requested):
Please run ruff format, otherwise LGTM

jelmer (approved):

PR #424, opened 2025-02-21
Remove support for Python 3.8 + deprecated methods and classes

stephenfin:
- **Drop support for Python 3.8**
- **Remove remnants of setuptools**
- **Remove deprecated classes, methods**
- **Resolve deprecation warning**


jelmer (changes requested):
LGTM, but looks like there are some test failures.

jelmer:
@stephenfin looks like it just needs to have "ruff format" run on it

testtools/__init__.py
```diff
@@ -109,21 +109,8 @@
...
-    except ModuleNotFoundError:
-        # The user is probably trying to run this without having installed
-        # the package, so complain.
-        raise RuntimeError(
-            "Testtools is not correctly installed. Please install it with pip."
-        )
+    from ._version import __version__, version
+except ModuleNotFoundError:
+    # package is not installed
+    version = '0.0.0'
```
jelmer, line 115: This is actually a functional change, breaking version strings for non-installed versions. Would it be possible to avoid this change and perhaps move it to a separate PR?

stephenfin: > breaking version strings for non-installed versions

Is that a real concern? tox and nox both install the package by default, and I expect most/all people to be running tests this way.

As a separate aside, this previously only worked if `setuptools_scm` was installed, which it generally won't be nowadays thanks to build isolation, and if you're working within the context of the git repo rather than e.g. a tarball. If the former isn't true, you'll get an exception. If the latter wasn't true, you'd get the same version you have here. It all feels a bit strange, especially when we're no longer using setuptools itself and when we expect people to be installing packages (and for those that insist on not doing so, a consistent `0.0.0` seems a perfectly reasonable value to return).

Would it be okay to drop this if we add a Changelog entry?

jelmer: I'd rather have that conversation separately of dropping python 3.8 support, rather than as a side-effect of this PR.

Personally I currently rely on this working for testtools - specifically also because breezy checks the testtools version string.

stephenfin: Would it be possible to add this logic to the caller instead so?

```python
if testtools.__version__ == (0, 0, 0):
    ...  # do other thing
```

This is also being flagged in https://github.com/testing-cabal/testscenarios/pull/11, and I think it's a rather poor solution to have here, let alone to copy to other projects. I'd rather we found an alternative if at all possible.

jelmer: How would we check the version number in that case? I'd be open to using something other than setuptools_scm, but defaulting to a (0, 0, 0) tuple seems like a step back.

This is also the behaviour that pbr gave us previously, so all of the testing-cabal projects used to have this behaviour.

stephenfin: I've copied some logic I wrote a while back for [Patchwork](https://github.com/getpatchwork/patchwork/blob/main/patchwork/version.py) to get this directly from Git, avoiding the need to use `setuptools_scm` for this side-effect. Hopefully that works :crossed_fingers:

jelmer: Thanks!

PR #457, opened 2025-07-26
Add typing

jelmer:


stephenfin:
@jelmer I am looking for this also. I only rebased this and resolve the merge conflicts, but I have a few extra fixes. If it's okay with you, I plan to push these on top of yours.

PR #490, opened 2025-12-17
Add typing

stephenfin:
This is an alternative to #457. I have pushed it to a separate PR since the changes are fairly substantial, though authorship of commits is retained.


jelmer:
@stephenfin Thanks!

PR #575, opened 2026-02-28
Add pre-commit

stephenfin:
Seeing as I keep forgetting to run ruff when submitting PRs :sweat_smile:


testtools/run.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
```
jelmer: Why add the shebang here? This shouldn't be invoked as ./testtools/run.py

scripts/all-pythons
```diff
@@ -72,22 +69,24 @@ def run_for_python(version, result, tests):
...
     return datetime.now(tz=iso8601.Utc())
 
 
-
-if __name__ == '__main__':
+if __name__ == "__main__":
     sys.path.append(ROOT)
     result = TestProtocolClient(sys.stdout)
-    for version in '3.10 3.11 3.12 3.13 3.14'.split():
+    for version in "3.10 3.11 3.12 3.13 3.14".split():
```
jelmer: not a hard blocker, but is there some way for avoiding another list of supported pythons here?

jelmer (changes requested):

tox.ini
```diff
@@ -12,10 +12,9 @@ commands =
 
 [testenv:ruff]
 deps =
-  ruff
+  pre-commit
 commands =
-  ruff check --fix .
-  ruff format .
+  pre-commit run -a
```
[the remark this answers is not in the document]
stephenfin: The concern with not doing this is that you end with two linter configurations which may subtly differ. This usually means someone who uses `pre-commit` will end up getting a whole load of fixes on unrelated code because others haven't run it for X time.

More to the point, I don't think there's really a dependency here. It's not needed at runtime. Distro packagers don't need to worry about. Devs don't need to even install it on their system since they can just rely on tox to do so. (I also wouldn't call tox a dependency for similar reasoning)

scripts/all-pythons
stephenfin: Oh I didn't add this script: I just ran the linter over it and ruff fixed the quotes (our existing tox configuration was missing this directory).

idk what this even does. I'll see if I can remove if outright...

stephenfin:
I don't know when I'll get back to this, and I'll likely start from scratch when I do. Let's close this.

PR #595, opened 2026-03-26
Vendor LoggingFixture and make fixtures, testresources hard requirements for testing

stephenfin:
Split out from https://github.com/testing-cabal/testtools/pull/593.

Note that we are not making these hard dependencies at runtime, only for the test suite (which is no longer packaged).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ requires-python = ">=3.10"
 Homepage = "https://github.com/testing-cabal/testtools"
 
 [project.optional-dependencies]
-test = ["testscenarios", "testresources"]
+test = ["fixtures", "testscenarios", "testresources"]
 twisted = ["Twisted", "fixtures"]
 dev = [
     "ruff==0.15.10",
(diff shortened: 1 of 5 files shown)


jelmer:
If we're going to add a dependency on fixtures, any reason we can't import LoggingFixture from there?

stephenfin:
> If we're going to add a dependency on fixtures, any reason we can't import LoggingFixture from there?

No reason, beyond the fact that there are a few releases in the wild with the fixture missing (we need both https://github.com/testing-cabal/fixtures/commit/53f7068b2b346dffeed71c3ee66d80f25dfaee00 and https://github.com/testing-cabal/fixtures/commit/f3f574a62628dc99b952da7a96a63637386961ec included to fix it), plus the `fixtures.tests` module we were relying on is now deprecated so we'd need a conditional import or would have to bump the minimum version of fixtures. Given all of this, it seemed easier to just vendor and continue working with all versions of fixtures

jelmer (approved):

PR #630, opened 2026-09-08
Add subTest support for testtools.TestCase

CyrilRoelandteNovance:
Commit 267c0be80d6de80242ff427c4b3e5d15086d64f7 added support for
unittest.TestCase.subTest. This meant the following code:

    import unittest

    class TestSubTests(unittest.TestCase):
        def test_even_numbers(self):
            for i in range(5):
                with self.subTest(i=i):
                    self.assertEqual(i % 2, 0)

would list the exact values of "i" for which the test failed:

$ stestr run test_example_subtest 2>&1 | grep ^Captured
Captured traceback (i=1):
Captured traceback (i=3):
Captured traceback (i=1):
Captured traceback (i=3):

But the following code:

    import testtools

    class TestSubTests(testtools.TestCase):
        def test_even_numbers(self):
            for i in range(5):
                with self.subTest(i=i):
                    self.assertEqual(i % 2, 0)

would not produce a similar output:

$ stestr run test_example_subtest_testtools 2>&1 | grep ^Captured
Captured traceback:
Captured traceback:

This commit fixes this so that we get the same output whether we use
unittest.TestCase.subTest or testtools.TestCase.subTest.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>
Signed-off-by: Cyril Roelandt <cyril@redhat.com>


testtools/testcase.py
```diff
@@ -99,6 +100,40 @@ class _ExpectedFailure(Exception):
     """
 
 
+_subtest_msg_sentinel = object()
+
+
+class _SubTestInfo:
```
jelmer: _SubTestInfo is a bare object, but it's passed to result.addSubTest() where a unittest.TestCase is expected. Stdlib's TextTestResult.addSubTest calls getDescription(subtest) ->
subtest.shortDescription(), so running any testtools test that uses subTest under a plain unittest.TextTestRunner aborts the run:

AttributeError: '_SubTestInfo' object has no attribute 'shortDescription'

Reproduced with unittest.TextTestRunner(verbosity=2) on a case with four subtests, two failing.

testtools/runtest.py
```diff
@@ -184,6 +184,9 @@ def _run_core(self) -> None:
                     if getattr(self.case, "force_failure", None):
                         self._run_user(_raise_force_fail_error)
                         failed = True
+                    for subtest, err in getattr(self.case, "_subtest_failures", ()):
+                        self.result.addSubTest(self.case, subtest, err)
```
jelmer: If the result object doesn't implement addSubTest, the resulting AttributeError is caught and discarded by the surrounding _run_user machinery, and a test with genuinely failing subtests reports as passing:

events: ['startTest', 'stopTest']
was_successful: True

Silently converting failures into successes is the worst available failure mode here — worth handling the missing-method case explicitly, or guaranteeing addSubTest is present on the decorated result.

testtools/testcase.py
```diff
@@ -887,6 +923,19 @@ def _report_traceback(
...
 
+    @contextlib.contextmanager
+    def subTest(
+        self, msg: object = _subtest_msg_sentinel, **params: Any
+    ) -> Iterator[None]:
+        """Return a context manager for a subTest."""
+        subtest = _SubTestInfo(self, msg, params)
+        try:
+            yield
+        except Exception:
```
jelmer, line 940: This catches SkipTest, so self.skipTest(...) inside a subTest is recorded as an error rather than a skip. Against stdlib, for a case that skips one subtest and fails another.

testtools/testcase.py
```diff
@@ -887,6 +923,19 @@ def _report_traceback(
             ),
         )
 
+    @contextlib.contextmanager
+    def subTest(
+        self, msg: object = _subtest_msg_sentinel, **params: Any
+    ) -> Iterator[None]:
+        """Return a context manager for a subTest."""
+        subtest = _SubTestInfo(self, msg, params)
```
jelmer: There's no nesting stack, so nested subTests lose the outer parameters.

jelmer (changes requested):

testtools/testcase.py
CyrilRoelandteNovance: OK, I ended up making this class inherit from unittest.TestCase so it has shortDescription and every method/attribute one might expect to be able to use.

testtools/runtest.py
CyrilRoelandteNovance: OK, I am now calling addFailure() if addSubTest() does not exist. I think we can expect addFailure to be implemented?

testtools/testcase.py
CyrilRoelandteNovance: OK I caught SkipTest and reraised the exception.

testtools/testcase.py
CyrilRoelandteNovance: I did not even think about that use case! I am now keeping a list of parameters.

CyrilRoelandteNovance:
@jelmer Thanks for the review, the new patchset should address all your comments.

testtools/testcase.py
```diff
@@ -887,6 +925,25 @@ def _report_traceback(
...
+    def subTest(
+        self, msg: object = _subtest_msg_sentinel, **params: Any
+    ) -> Iterator[None]:
+        """Return a context manager for a subTest."""
+        merged_params = {**self._subtest_params, **params}
+        subtest = SubTest(self, msg, merged_params)
+        old_params, self._subtest_params = self._subtest_params, merged_params
+        try:
+            yield
+        except SkipTest:
```
jelmer, line 938: is raising here correct? I think we should record this subtest as skipped but still run the other ones

testtools/runtest.py
```diff
@@ -184,6 +184,12 @@ def _run_core(self) -> None:
                     if getattr(self.case, "force_failure", None):
                         self._run_user(_raise_force_fail_error)
                         failed = True
+                    for subtest, err in getattr(self.case, "_subtest_failures", ()):
+                        try:
+                            self.result.addSubTest(self.case, subtest, err)
+                        except AttributeError:
```
jelmer, line 190: this also catches AttributeErrors from within addSubTest, not just absence of addSubTest itself

CyrilRoelandteNovance: Good catch. I switched to testing whether the method exists before calling it.

testtools/testcase.py
CyrilRoelandteNovance: Indeed, I should have logged the reason for skipping and moved on to the next subtests. This is fixed in the new patchset.

jelmer (approved):

